PR #20, opened 2015-12-30
Fix sprinting section

### LINKED ISSUE #16: Sprinting section needs breaking up
It's a bit wall-of-text, it could do with breaking into headings and probably a bit more detail.

onishiweb:
Fixes #16

Update the sprinting section to break it out into better sections and add a bit more detail and cross-links to other areas of the playbook.

### DIFF
--- a/playbook.md
+++ b/playbook.md
@@ -136,13 +136,21 @@ This isn't to say that we require clients to participate in everything. Clients 
 
 
 ### Sprinting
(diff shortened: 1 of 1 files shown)


harry-m:
There've been quite a few conversations since this PR and I think this section is in better shape now. It also seems to contain most of the changes from this PR, though I'm not sure how! So I'll close this.

PR #25, opened 2016-05-24
Branching, stories, pull requests and code reviews

tahb:
I've written this based on how the Rails team has been working with the hope that everyone will add their thoughts and we will reach a consensus! 

* Lifecycle of a story - a new ideal as a result of a conversation we've had about how we'd like to work and a dependent move away from Pivotal Tracker so we can have custom columns/state.


tahb:
Rebased and resolved the conflicts.

Ready to merge @harry-m if you're happy with it.

PR #29, opened 2016-06-15
Detail our own commit conventions

tahb:
* previously we simply pointed to angular which nobody read
* the rails team has evolved this set of basic rules which has worked well


tommorris:
Nitpick: `well-bounded` not `well bounded`.

Otherwise 👍.

tahb:
Thanks @tommorris, I've added that.

PR #31, opened 2016-06-21
Starting a style guide

tommorris:


harry-m:
Good addition, I think this also needs a section on how we write things.

PR #34, opened 2016-08-10
Add note about using "Resolves:" in commit messages

mallorydxw:
I've been using a mixture of `Ticket: https://dxw.zendesk.com/...`, `Story: https://www.pivotaltracker.com/...` and `Fixes: #123` (for GitHub issues) for a while now, but I'd like to standardise, and it would be good to get some consistency on this.

On the subject of commit squashing - I used to make very small commits, but as I've started to use TDD more I've gravitated to larger commits, and it seems like it's easier to write a detailed description for a monolithic commit than it is for microcommits. It also has a side-effect of making rebasing easier which makes my life easier, and will probably be much much easier for the less experienced git users.

### DIFF
--- a/playbook.md
+++ b/playbook.md
@@ -201,10 +201,16 @@ Every project will grow and with it the number of decision making processes we g
 
 Recording what these are is extremely helpful to enable anybody working on the project now and in the future to make informed decisions by reading the commit history.
 
+After the description there should be a line that starts with `Fixes:` followed by a link to the Zendesk ticket, Pivotal story, or GitHub issue (do not use `#123` instead of providing the full URL - who knows if that's a GitHub issue or a Zendesk ticket). This should not be used as a substitute for a description - the description will last as long as the repository, the link to Zendesk will last as long as Zendesk is profitable. Naming the branch `hotfix/123` is not a substitute for including this line as branch names are ephemeral and only recorded on the merge commit (and there's no easy way to find a merge commit from a commit).
+
+If a single ticket/story is being completed, a squashed commit is preferred over multiple commits.
+
 
 ```
 # Poor
 Fixed the form
+
+Fixes: #123
 
 # Great
 (fix): Food order form works with special characters
@@ -212,6 +218,8 @@ Fixed the form
 * Previously the form failed to send if the starter value included any of the following characters: !@£$%^&*().:;'
 * Users should be able to use these characters and after a short investigation I found this new bug in PHP: <link>
 * As a short term workaround until the bug is fixed I am stripping out any of these characters if they are present.
+
+Fixes: https://dxw.zendesk.com/agent/tickets/0000
 ```
 
 #### Code style
--- a/playbook.md
+++ b/playbook.md
@@ -201,16 +201,18 @@ Every project will grow and with it the number of decision making processes we g
 
 Recording what these are is extremely helpful to enable anybody working on the project now and in the future to make informed decisions by reading the commit history.
 
-After the description there should be a line that starts with `Fixes:` followed by a link to the Zendesk ticket, Pivotal story, or GitHub issue (do not use `#123` instead of providing the full URL - who knows if that's a GitHub issue or a Zendesk ticket). This should not be used as a substitute for a description - the description will last as long as the repository, the link to Zendesk will last as long as Zendesk is profitable. Naming the branch `hotfix/123` is not a substitute for including this line as branch names are ephemeral and only recorded on the merge commit (and there's no easy way to find a merge commit from a commit).
+After the description there should be a line that starts with `Resolves:` followed by a link to the Zendesk ticket, Pivotal story, or GitHub issue (do not use `#123` instead of providing the full URL - who knows if that's a GitHub issue or a Zendesk ticket). This should not be used as a substitute for a description - the description will last as long as the repository, the link to Zendesk will last as long as Zendesk is profitable. Naming the branch `hotfix/123` is not a substitute for including this line as branch names are ephemeral and only recorded on the merge commit (and there's no easy way to find a merge commit from a commit).
 
-If a single ticket/story is being completed, a squashed commit is preferred over multiple commits.
+Other links relevant to the commit should be added as lines beginning with `Link:`.
+
+If a single ticket/story is being completed, a [squashed commit](https://robots.thoughtbot.com/git-interactive-rebase-squash-amend-rewriting-history#squash-commits-together) is preferred over multiple commits.
 
 
 ```
 # Poor
 Fixed the form
 
-Fixes: #123
+Resolves: #123
 
 # Great
 (fix): Food order form works with special characters
@@ -219,7 +221,8 @@ Fixes: #123
 * Users should be able to use these characters and after a short investigation I found this new bug in PHP: <link>
 * As a short term workaround until the bug is fixed I am stripping out any of these characters if they are present.
 
-Fixes: https://dxw.zendesk.com/agent/tickets/0000
+Resolves: https://dxw.zendesk.com/agent/tickets/0000
+Link: http://grimm.blog/solution-to-ie7-render-bug
 ```
 
 #### Code style


tomdxw:
Thoughts?

@dgmstuart @harry-m @tahb @tommorris (People discussing commit messages in issues, and people whose names showed up when I ran `git blame` on the commit messages section).

tommorris:
Regarding commit squashing, we should probably include a link to something like [this post](https://robots.thoughtbot.com/git-interactive-rebase-squash-amend-rewriting-history) in order to explain to new dxw-ers how to engage in Git rebasery.

I also doubt the wisdom of monolithic commits "making rebasing easier". I frequently make a dozen little microcommits then roll them up into logical commits using `git rebase -i`. Like learning a new language or playing the violin or whatever, there's not really _any_ way to make that easy. It requires understanding the theory and practicing it.

tahb:
@tomdxw I agree with the idea of more formally adding the links into the description for all types of commit because it decouples us from the software we use. As you say, the ID we include in branch names could relate to Pivotal, Zendesk, Trello or Alibaba 👍  

I would however be keen to make sure we also don't use this as a replacement for the ID in each of the branch names. It would be good to know if I'm being a special snowflake on this but in my workflow it makes it very easy to scan a list of remotes for branches connected to a particular story or ticket (especially when there are multiple). Is this something others value?

Otherwise I only have an idea on how to expand upon your point. Instead of having the `fixes:` prefix as a thing (which to me implies ticket not a story) how about simply scratching it to encourage the inclusion of **any and all** relevant links. I feel that recording as many relevant pieces of internet alongside the parent story/ticket should be something we try to encourage as it makes it much faster for another developer to boot their brain up around a piece of code.

Another great example could be:
```
# Great
(fix): Receipt page now renders content on IE7

* A user found an issue where the receipt page didn't render any content at all when viewed in IE7.
* I found a blog post by Mr Grimm where he has found this issue and details his preferred solution 
of including IE7 logic in the layout to catch these browsers, which I agree with as a clean and
isolated way to address this issue.

http://dxw.zendesk/123
http://grimm.blog/solution-to-ie7-render-bug
```

tomdxw:
> I would however be keen to make sure we also don't use this as a replacement for the ID in each of the branch names.

I like that convention. I see no reason to change that.

>  Instead of having the `fixes:` prefix as a thing (which to me implies ticket not a story) how about simply scratching it to encourage the inclusion of any and all relevant links.

I think we should use the lines which look like HTTP headers. It's a very common convention. For example, it's used in:

- linux: https://github.com/torvalds/linux/commit/57dea93ac42d341e1fe902528b348ef6763fa485
- git: https://github.com/git/git/commit/27b0ea4038f5976bee9d183ccbc868e9fcd9825e
- GitHub uses `Fixes:` to auto-close issues
- Gerrit can be configured to reject merge requests that don't have the appropriate `Signed-off-by:` line

"Resolves" sounds fairly neutral as to whether the thing is a bug or a feature. So how about something like this:

```
Resolves: https://dxw.zendesk/agent/tickets/123
Link: http://grimm.blog/solution-to-ie7-render-bug
Link: https://css-tricks.com/ie-css-bugs-thatll-get-you-every-time/
Link: https://dxw.zendesk/agent/tickets/456
```

This way it's clear that this commit fixes the issue raised in #123, but ticket #456 merely holds some relevant information.

tomdxw:
> Regarding commit squashing, we should probably include a link to something like this post in order to explain to new dxw-ers how to engage in Git rebasery.

Added that link. Thanks.

tahb:
> GitHub uses Fixes: to auto-close issues

That I didn't know!

Naming things is a pain I know but to me `resolves:` implies there was a problem which is good for tickets but `delivers:` could be a better fit for stories?

```
fixes: <ticket>
delivers: <story>
```

tahb:
It would be good to talk about encouraging large commits over talky @tomdxw. This is something we have always been actively against and I think for good reason.

harry-m:
I agree with @tahb - happy to discuss but I don't think I like the idea of big commits either. For reasons:

* It's harder to isolate and revert a specific change (like a bit of refactoring in the middle of a story)
* It makes explanations of changes much less connected to the changes themselves. A big commit will change lots of things and each of those things might need a bit of explanation of its own. It's easier to see what explanation applies to what if the commits are smaller

That said, I'm not sure I'd immediately identify "microcommits" as being good either. I think the granularity should be appropriate for the task at hand.

dgmstuart:
@tahb I massively disagree that `fixes` or `resolves` implies a ticket and not a story. I don't think we have a 'ticket = bugfixes' and 'story =  feature' split, particularly now that we're continuously sprinting.

I thought the way we've been working with bugs is: 
1. create a ticket
2. investigate the cause and work out what needs to be done
3a. if it's quick, just do it
3b. if not, create a story

Or if we know what needs to be done, and it's a non-trivial amount of work, then we'll just skip straight to 3b.

Also I feel like it's legit to create small 'fix' commits at the beginning of a PR when small low-level bugs are discovered in the course of development.

harry-m:
@dgmstuart Does small 'fix' commits in a bigger story contradict the section on pull requests? Or should we make this less absolute?

> A good pull request should:
> // snip
> * Focus on the single problem at hand. Including other changes will make it much harder to merge

https://github.com/dxw/playbook/blob/master/playbook.md#pull-requests

I think fixes and delivers could each apply to both tickets and stories - not sure that it's important that we make the distinction, really?

@tahb any view on that?

@tomdxw any view on big commits vs granular ones?

The more I think about it, the more I think commit squashing is a bad idea. It breaks the whole idea of using commit messages and descriptions to document changes, does it not?

tahb:
Incoming wall, sorry!

> I thought the way we've been working with bugs is: 
1. create a ticket
2. investigate the cause and work out what needs to be done
3a. if it's quick, just do it
3b. if not, create a story

This has changed a lot recently but this is how I currently think about bugs:

Bugs that **directly relate to the story** I'm working on I fix at the time and commit with (fix) to my pull request before adding the feature commits.

Bugs not related to the story mean I need to track it somewhere, and that depends on its type. If it's related to application code then I write a story with a need that isn't currently being met and if it's to do with the platform then I create a ticket. 

In either case I don't think they should automatically become a ticket for the retained developer as this will lead to an overload of application bugs and no work on supporting the platform. Leading to it becoming stale again and so on. Arguably the retainer should triage front-line tickets into stories as well.

> I think fixes and delivers could each apply to both tickets and stories - not sure that it's important that we make the distinction, really?

Yeah I see, they both could be in different situations. Happy to go with `fixes:` for everything.

> The more I think about it, the more I think commit squashing is a bad idea. It breaks the whole idea of using commit messages and descriptions to document changes, does it not?

Agree. For the project there is more value in small, well documented commits than having a short term way to make rebasing easier.

I think if rebasing is hard, there could be other less detrimental ways to make collaboration easier. 

* pull requests that are concise and focused should rarely conflict dramatically with another developers PR. If there genuinely is a lot to focus on then this probably means the story is too complex and should be split out
* not rebasing branches that are in progress on top of `develop` little and often will mean that the eventual rebase will be naturally larger and with more chance of conflicting. Again it feels like this can only occur if the branch in progress is there for more than a few days, meaning a complex story

dgmstuart:
@tahb the reason we decided to create tickets for bugs where we don't know the root cause (just what the symptoms are), is because it's impossible to estimate when you don't know what work needs to be done. The idea was that this sort of triage should be done in support time. 

Has that changed, or are we actually agreeing on that point and misunderstanding each other?

dgmstuart:
@harry-m rest assured that routine commit squashing is something that many well-respected teams and open source projects do. I'm not saying we should do it here, but this is one of those things where we can investigate what other people are doing (and why): we don't need to speculate.

dgmstuart:
@tahb
> Agree. For the project there is more value in small, well documented commits than having a short term way to make rebasing easier.

I'm missing some context here: in what way is rebasing hard? Or at least: in what way is it harder than merging?

dgmstuart:
The reason I put small bugs in at the beginning of a PR, *even if they're not related to the story* is because there is significant overhead in creating a new PR and getting it reviewed, compared to just adding it into another PR. 

Maybe that overhead is a smell, but I'm not sure I see a way to address it.

harry-m:
Ok. I think we are now discussing too many things in one place.

This PR is about using "fixes" in commit messages and I think we are agreed that it's fine to do so and beneficial where you are referencing an issue. Otherwise I think we have agreed that we'll use the existing convention of (fix), (feature) etc with a link to the work in whatever the relevant tracking tool is.

Does everyone agree on that?

I suggest that we take the rest of this conversation offline. To be discussed: granularity of commits, commit squashing, issues around rebasing and conflicts, what to do with bugs that are unrelated to the purpose of the PR. @imagecircus please can you arrange another all-dev chat for that and anything else that's pressing.

harry-m:
@imagecircus Can we get something booked in next week?

imagecircus:
We'll discuss this in next weeks dev chat.

tomdxw:
I think `Resolves:` is better than `Fixed:` because it doesn't imply that the thing in the link is a bug.

harry-m:
Was this discussed?

dgmstuart:
@harry-m the next dev chat hasn't happened yet! It's this Friday

tahb:
There was a dev chat but it didn't cover Git. It will hopefully be covered this Friday.

tomdxw:
I don't think we discussed this.

tahb:
We discussed in the dev chat (28th October) that it would be great to avoid the repetition of adding the same link to each commit, given that there is usually several for each story. 

We should already be tying each pull request to it's associated card in Trello in some way, whether that be including the ID in the title or manually adding (the GitHub power-up makes this a breeze) a link to the description.

The problem comes when we look back in time to understand why an individual commit was made and what is was connected to. GitHub provide an way to do this but frustratingly GitLab doesn't, at least from my playing around with it.

tomdxw:
> it would be great to avoid the repetition of adding the same link to each commit, given that there is usually several for each story.

Commit squashing fixes that. I never used to do commit squashing but now I'm a fan of it.

> GitHub provide an way to do this

`tig` doesn't provide a link to the GitHub PR. Links are better because they work with all tools.

> but frustratingly GitLab doesn't, at least from my playing around with it.

The merge commit message contains a reference to the MR:

https://git.govpress.com/gds-campaigns/app/commit/c143fb1751ee19a241320c078962dee14703a400

It's better than nothing, but it's not ideal because a) you need to navigate forwards in the history to get to it, and b) it's not a URL so there's no easy way to navigate to the MR when using any git tool except GitLab.

dgmstuart:
This seems stale

All in favour of closing?
@tahb @tomdxw @harry-m (@imagecircus ?)

PR #48, opened 2016-10-04
Ticket principles: making clients feel informed and in control

harry-m:
I think this is something that our company values cover well, but is notably missing from the ticket principles. We should talk about it and add something.

### DIFF
--- a/playbook.md
+++ b/playbook.md
@@ -396,7 +396,7 @@ Of course, every client would like us to go the extra mile to solve their proble
 
 While we do everything we can to make sure the client is happy with our solution, we are also mindful of what's practical. We don't do serious bits of development work on tickets, or trial new approaches. We don't play with new tools or sink hours into interesting bugs. We set those things aside, and do them later.
 
-The main purpose of a ticket is to take some action that solves the problem quickly, particularly when the availability or functionality of the site is compromised. However, a bad solution implemented quickly is rarely preferable to a good solution implemented slightly less quickly. Time should not be the deciding factor if the quick fix will make our or the client's life significantly more difficult, or if a small investment of time up front will save time in future.
+The main purpose of a ticket is to take some action that solves the problem quickly. However, a bad solution implemented immediately is rarely preferable to a good solution implemented slightly less quickly. Time should not be the deciding factor if the quick fix will make our or the client's life significantly more difficult, or if a small investment of time up front will save time in future.
 
 #### Make sure clients feel informed and in control
 For many of our clients, the helpdesk can feel intimidating: especially when they are not technical, or are not sure what they are entitled to ask for as part of the service they receive.
--- a/playbook.md
+++ b/playbook.md
@@ -289,13 +289,13 @@ On WordPress projects, we [deploy by pushing commits](https://github.com/dxw/pla
 
 
 ### Managing delivery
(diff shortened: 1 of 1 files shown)


playbook.md
```diff
@@ -396,6 +398,15 @@ While we do everything we can to make sure the client is happy with our solution
 
 The main purpose of a ticket is to take some action that solves the problem, as quickly as possible. Generally speaking, we do the most time-efficient thing that we can. Of several potential solutions that solve the problem, assuming none is a bad one, we do the one which can be implemented the most quickly.
 
+#### Make sure clients feel informed and in control
+For many of our clients, the helpdesk can feel intimidating: especially when they are not technical, or are not sure what they are entitled to ask for as part of the service they receive.
+
+It is important that we make the helpdesk experience a [positive](#positive) one. Tickets will not always explain an issue in a way that is clear to us, and clients will not always ask for the right thing. 
+
+It is our responsibility to gain an accurate understanding of the issue, and to help the client to feel more informed in the process. Where there are many possible solutions, it is also our responsibility to present those options to the client, such that they are able to understand each option's strengths and make an informed decision.
```
harry-m: I agree. Could you guys like to amend to reflect that?

RobjS: @harry-m @Stretch96 I've had a go at amending the wording around timing to make it clear that's not the only factor to consider (see l.399 - other changes are just my editor removing trailing spaces).

harry-m: I like the change, but I think it is also worth having some words in there to make it clear that working on a ticket is not the time to shave yaks, dive down rabbit holes or gold-plate things. 

Is it worth adding some words to the effect of: "fix things properly, but don't go over the top", or am I worrying about an edge case?

RobjS: Doesn't the preceding para cover that?

`While we do everything we can to make sure the client is happy with our solution, we are also mindful of what's practical. We don't do serious bits of development work on tickets, or trial new approaches. We don't play with new tools or sink hours into interesting bugs. We set those things aside, and do them later.`

harry-m: Ah. Hah. Yes, it totally does. I didn't read that bit because it wasn't green :)

CristinaRO:
This PR hasn't had any discussion in over a year, so we've decided to close it (Dev forum Oct 19 2018).

harry-m:
Hmm, I am not sure about zapping this one. I hadn't realised it wasn't merged (itself not a good thing). I think this bit could use a discussion, and if the team are in agreement with some form of it, I think we should include it:

> #### Make sure clients feel informed and in control
> For many of our clients, the helpdesk can feel intimidating: especially when they are not technical, or are not sure what they are entitled to ask for as part of the service they receive.
> It is important that we make the helpdesk experience a [positive](#positive) one. Tickets will not always explain an issue in a way that is clear to us, and clients will not always ask for the right thing.
> It is our responsibility to gain an accurate understanding of the issue, and to help the client to feel more informed in the process. Where there are many possible solutions, it is also our responsibility to present those options to the client, such that they are able to understand each option's strengths and make an informed decision.
> Much of the [Involvement](#involvement) section applies to ticket work too. In particular, that we'll know we're doing this well when our clients feel like they understand what is going on and are in control of their ticket's outcome.

PR #51, opened 2016-10-07
Describe each state of a story

tahb:
* These align with the column names we are using in Trello
* Describe what must be true of each story for the state to change from a development perspective in greater detail
* My editor seems to have picked up some white space on the end of two unrelated lines, removing this here to save the next person

### DIFF
--- a/playbook.md
+++ b/playbook.md
@@ -178,17 +178,42 @@ Each story will also contain a list of acceptance criteria. Acceptance criteria 
 For more information about writing good stories, read User Stories Applied by Mike Cohn. There is a copy of this book in the dxw library.
 
 
-#### Lifecycle of a story
-There are several states that a story has to go through in order to be deployed to production. We use tracking tools and physical story boards to keep track of which stage a given story is in.
+## Lifecycle of a story
+There are several states that a story has to go through in order to be deployed to production. We use tracking tools and physical story boards to keep track of which state a given story is in.
 
-* **Started**: A developer has started working on the story, making changes to the product to ensure that each acceptance criterion is met
-* **Code review**: The first developer has finished their work, and is ready for a second developer to [review](#Code reviews) the technical aspects of the story in a [pull request](#Pull request)
-* **Acceptance review**: The two developers have finished the review and the story is now merged, and ready for the delivery manager to review against its acceptance criteria
-* **Client review**: The delivery manager has confirmed that the story is done, and it's ready for a client to review
-* **Ready to deploy**: The client has accepted the story and it's ready to deploy
-* **Deployed**: The story is deployed to production
+### States
 
-#### Branching
+After a user need has been identified, a story is written to capture it and how we propose to meet that need with acceptance critera. Once it's written and prioritised as part of a sprint planning it will be ready to start.
+
+#### Unstarted
+- hasn't been started or isn't being actively developed yet and can be changed
+
+#### In progress
+- a developer or a pair are currently working on meeting the acceptance criteria
+
+#### Awaiting code review
+- a pull request has been opened
+- another developer is required to review the codes quality and merge into develop
+- this developer also tests the functionality with the aim of catching the majority of any bugs as early as possible and feedback them back onto the pull request
+
+#### Awaiting delivery manager review
+- the work to fulfill the story is now on staging
+- a delivery manager is required to review the product and ensure that it's ready for the client to sign off
+- a show and tell from the developer may be needed to demonstrate stories that are hard to test
+
+#### Awaiting client review
+- a delivery manager has explained what is ready and if necessary how to review it
+- the client can use Staging to see that the acceptance criteria have been met
+- based on this mark the story as accepted or rejected and include any details of which criteria hasn't been met
+
+#### Accepted
+- the client has accepted the story as the acceptance criteria on the story are met and it is ready to be deployed to production
+
+#### Deployed
+- the new story is ready to be used by the users
+- a deployment has taken place (`develop` has been merged into `master`)
+
+## Branching
 When we start any new piece of work - whether story, bug fix, or chore - we always start by creating a new branch based off the repository's `develop` branch.
 
 We have a naming convention for branches, as follows:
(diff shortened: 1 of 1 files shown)


playbook.md
```diff
@@ -178,17 +178,42 @@ Each story will also contain a list of acceptance criteria. Acceptance criteria
...
+- a delivery manager has explained what is ready and if necessary how to review it
+- the client can use Staging to see that the acceptance criteria have been met
+- based on this mark the story as accepted or rejected and include any details of which criteria hasn't been met
+
+#### Accepted
+- the client has accepted the story as the acceptance criteria on the story are met and it is ready to be deployed to production
+
+#### Deployed
+- the new story is ready to be used by the users
+- a deployment has taken place (`develop` has been merged into `master`)
```
tomdxw: How is `develop` merged into `master`?

playbook.md
```diff
@@ -178,17 +178,42 @@ Each story will also contain a list of acceptance criteria. Acceptance criteria
...
+
+#### Unstarted
+- hasn't been started or isn't being actively developed yet and can be changed
+
+#### In progress
+- a developer or a pair are currently working on meeting the acceptance criteria
+
+#### Awaiting code review
+- a pull request has been opened
+- another developer is required to review the codes quality and merge into develop
```
tomdxw: "code's"

playbook.md
```diff
@@ -178,17 +178,42 @@ Each story will also contain a list of acceptance criteria. Acceptance criteria
...
+- this developer also tests the functionality with the aim of catching the majority of any bugs as early as possible and feedback them back onto the pull request
```
magdamaciaszczyk: This will automatically increase ticket time because of the amount of WP projects we have and the fact that some projects are not regularly worked on

playbook.md
```diff
@@ -309,7 +334,7 @@ By the end of this session, the full team should be confident that there is enou
 TODO
 
 #### Retrospectives
-At the end of every sprint, the delivery manager facilitates a [retrospective](https://www.gov.uk/service-manual/agile/running-retrospectives.html) where the team can discuss how the sprint went. These sessions are attended by all the people involved in delivering the sprint along with the client team. 
+At the end of every sprint, the delivery manager facilitates a [retrospective](https://www.gov.uk/service-manual/agile/running-retrospectives.html) where the team can discuss how the sprint went. These sessions are attended by all the people involved in delivering the sprint along with the client team.
```
magdamaciaszczyk: How this should be handled on WP projects?

playbook.md
tahb: To date manual merges have worked well. We can have confidence in them as everything on Develop has to have gone through a PR and CI. What are your thoughts?

playbook.md
tahb: Magda this is appearing in this pull request as my editor removed a tiny bit of whitespace so this is nothing new that I'm trying to add 😃

That said there should be retrospectives at the end of every project's sprint, if this isn't the case definitely bring it up with a DM!

playbook.md
magdamaciaszczyk: Of course. I agree that it's a good approach moving forward as we grow and more people are involved in each project.

playbook.md
magdamaciaszczyk: 👍

playbook.md
tomdxw: 1. Sometimes clients want changes to be on staging but they don't want those changes to be on production. So we'll just go through and manually revert all the changes except for the change we want to push? And then when we're done will revert the commit that reverted everything and push that to the develop branch? Is that correct?
2. And sometimes when we do this, the client may be in the middle of reviewing changes - and they'll come back to us and say "well it looked okay this morning but this seems to be broken right now - what's going on?". It's going to cause delays and confusion.
3. I thought one of the complaints about git env was the fact that it's a bit tricky to rebase and some people struggled to do it properly. But reverting large amounts of changes (which may have been merged in as separate branches on different days/weeks/months) just seems extremely complex. Can you explain how you would do it?

playbook.md
tomdxw: In most cases the client will be able to test that the feature is working on staging before being pushed to production. So I'm not convinced that this trade-off is worth it. But this is something I'm happy to experiment with.

One thing to note though:

@harry-m  told me that when we move to the 4-day work week, there will be one WP developer sprinting on tickets while all the other developers are sprinting on ... sprints. If people don't have time dedicated to tickets anymore, I would be concerned that people will get an SLA-breached ticket in the middle of a sprint and they don't really have time to get the site set-up and test it thoroughly, and we'll end up wasting an hour that a client has paid for to test something for another client.

playbook.md
tomdxw: Sometimes I find things hard to test locally, and instead I push the change to staging. How are we going to handle that now that a push to staging requires code review and testing?

playbook.md
magdamaciaszczyk: @tomdxw , good point. Not everyone from our WP team is able to do each dev ticket that is coming through so it might need to be assigned to someone else.

My main problem with testing tickets is that it also takes a while to set up a WP project if you haven't worked on a particular one for a while and I'm a bit worried waiting for review will take ages.

I think all these guidelines need to take into account the different nature of each of the sites we deal with and there should be a compromise somewhere.

playbook.md
magdamaciaszczyk: Exactly, this is my main concern with testing new features locally by other devs.

playbook.md
harry-m: @tomdxw that is a fair concern but we will arrange the process so that that won't happen. 

Current plan: if the support dev can't solve the issue it will be escalated to another dev to deal with on the following Friday, and the SLA will be configured to understand that and not to breach prematurely.

(hopefully this will not happen very often)

playbook.md
tahb: Regularly pushing code that is potentially volatile to an environment that is also used by clients and other members of the team will surely cause pain and confusion for everyone else. If there are exceptional circumstances where you're local environment cannot give you the confidence you need (it would be good to understand why this happens), then communicate this to the DM.

tomdxw: Okay, let's say I've disturbed @imagecircus or @staceydxw after they've gone home for the day, and they've approved me testing stuff on staging, how do I do it?

Do I force-push to the `develop` branch? Do I create a MR each time I want to push my changes? Or do I just merge them into `develop`? What if I need to push 5 times? Should I just commit directly to the `develop` branch? Otherwise I'd be creating a lot of branches and it would get annoying. And then do we need to keep those things in our commit history forever, and just use a `revert` commit to remove them?

Once I'm done testing how do I turn my commits into a branch to create a MR for? Do I cherry-pick? Do I `git checkout develop^ -- .`? What do you suggest as the easiest way to do it?

playbook.md
harry-m: When is the last time you had to exclude a change from production like this? How often is it happening? Can I have some examples?

tomdxw: > When is the last time you had to exclude a change from production like this?

When we were using a git flow-style workflow. That's _why_ we're using ENV branching now. This was one of the main reasons we made the switch.

> How often is it happening?

It isn't. Because we're using ENV branching now.

> Can I have some examples?

It will happen all the time. Example:

- Person A fixes a ticket, code review etc, and their changes are now on `develop` so that the client can review
- Person B fixes a ticket, code review etc, and their changes are now on `develop`
- The client responds to Person B and says it's good and they should deploy
- Now Person B needs to go through the changes that are on `develop` but not on `master`, they need to search for every change that was not part of this ticket and they need to do `git revert [commit hash]`
- And then they need to push that to `develop` and make sure everything's okay
- And then they need to merge `develop` into `master` (race conditions possible here)
- And then they need to go back to the `develop` branch and do `git revert` on every revert they've just made, and push again

That's my understanding of the process. But I don't know because the documentation doesn't say what to do in this situation.

playbook.md
tomdxw: > Is there any chance of a testing branch/environment that we could use for cases like this? One that isn't part of the main gitflow model (so master and develop remain untouched) but gives a remote environment that we can do anything we like with?

I've brought that up a few times. It's something we'd like to do. But we haven't made it a priority yet.

> I'm not really comfortable with the idea of a staging branch (whatever it's called) that is shared with the client, and amongst the team, that we can never be sure of the contents of. That seems to bring a whole load of problems of its own - what if one of us has pushed a fix there for client testing, and then someone else force pushes their own work on top to test it?

Generally speaking the advice when we were using ENV branching was that force-pushing should be used sparingly, and you should communicate with others when you do it.

tahb:
After this feedback and the other conversations I'm closing this PR and will replace it with something that I hope is easier to digest and includes answers to all your queries 👌

PR #54, opened 2016-10-14
Change all occurances of 'testing' to 'develop' and removed informati…

Stretch96:
Change all occurances of 'testing' to 'develop' and removed information about pushing to the old govpress.


tahb:
I've worked on stripping most of this file as there is now little distinction between WP and Rails projects: https://github.com/dxw/playbook/pull/56

If that is agreed upon, this can likely be closed as it incorporates the same change.

tahb:
I'm removing this file and am collecting it all into the new development-guide

PR #56, opened 2016-10-20
How we deploy our work is now generic for both teams

tahb:
- Now that branches and git workflow is the same for all projects
- This also removes references to Git ENV specific commands
- The only remaining differences between them are these services:
  - Remotes (GitLab or GitHub)
  - CI service (Chef or Heroku)


tahb:
Closing because this should live in the single development guide

PR #57, opened 2016-10-26
WIP: Add trello-workflow.md

tomdxw:


guides/trello-workflow.md
```diff
@@ -0,0 +1,53 @@
+# Trello Workflow
+
+New workflow. Uses Trello.
+
+## States
+
+- `backlog`
+- `in progress`
```
magdamaciaszczyk: Some boards use another list for Tickets. This might be useful when there is a sprint running and we want to see if there is any outstanding work outside of sprint. However, we can also leave these States flexible for each project because each is a bit different.

guides/trello-workflow.md
```diff
@@ -0,0 +1,53 @@
...
+TODO
+
+## Deploying code at any time
+
+Changes to deployed are in the `accepted` column. The `awaiting DM review` and `awaiting client review` columns must be empty.
+
+```
+% git fetch
+% git checkout master
+% git merge --no-ff origin/develop
```
magdamaciaszczyk: When I use `git merge origin/develop` it doesn't create an extra commit so we can have `master` and `develop` at the same point. Can we use this instead?

tomdxw:
Closing in favour of #58.

Issue #62, opened 2016-11-09
Style guide violates style guide (bullet points)


tomdxw:
https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#bullet-points-steps

Specifically:

- you always use a lead-in line
- you use lower case at the start of the bullet
- you don’t use more than one sentence per bullet point - use commas, dashes or semicolons to expand on an item

staceydxw:
I'm not sure I understand this? Are you suggesting that we use the GOV.UK style guide for the development guide?

tomdxw:
@staceydxw Sorry for not explaining properly. [Our style guide](https://github.com/dxw/playbook/blob/master/guides/style-guide.md#a-z) says "we also follow the GOV.UK style guide".

The README of this repo says "These documents are maintained by everyone at dxw, and it's important that their quality and tone is consistent." which one can assume means "use the styleguide for everything here".

So yes, everything in this repo needs to conform to the style guide.

PR #64, opened 2016-11-10
Update style guide to adhere to the style guide

tomdxw:
- Bullet points should start with a lower-case letter
- Add lead-ins for bullet points
- Remove e.g.
- Remove etc

Resolves: https://github.com/dxw/playbook/issues/62

### DIFF
--- a/guides/style-guide.md
+++ b/guides/style-guide.md
@@ -4,24 +4,29 @@
 
 In general:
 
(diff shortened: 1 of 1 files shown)


tomdxw:
I think the "Do this:" lead-ins are not very good. But I wasn't sure what else to use.

PR #70, opened 2016-12-20
Replace obfuscate

RobjS:
At the company healthcheck yesterday, there seemed to be a general wish to move away from the use of the term `obfuscate`. This is an attempt to replace the word with plainer English.

### DIFF
--- a/playbook.md
+++ b/playbook.md
@@ -30,7 +30,7 @@ We think that it's very important to have a talented team if we're going to succ
 We are helpful, useful and expert. We give practical and pragmatic advice to each other, and to our clients.
 
 #### Honest
-We are honest, trustworthy and straightforward. We give plain-spoken, frank, accurate feedback and advice, and we never mislead or obfuscate.
+We are honest, trustworthy and straightforward. We give plain-spoken, frank, accurate feedback and advice, and we never mislead or wilfully confuse.
 
 #### Positive
 We are positive, cheerful and supportive. Even in crisis, we stay optimistic. We [assume good faith](https://en.wikipedia.org/wiki/Wikipedia:Assume_good_faith) and offer constructive challenge.
--- a/playbook.md
+++ b/playbook.md
@@ -30,7 +30,7 @@ We think that it's very important to have a talented team if we're going to succ
 We are helpful, useful and expert. We give practical and pragmatic advice to each other, and to our clients.
 
 #### Honest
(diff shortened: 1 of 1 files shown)


playbook.md
```diff
@@ -30,7 +30,7 @@ We think that it's very important to have a talented team if we're going to succ
 We are helpful, useful and expert. We give practical and pragmatic advice to each other, and to our clients.
 
 #### Honest
-We are honest, trustworthy and straightforward. We give plain-spoken, frank, accurate feedback and advice, and we never mislead or obfuscate.
+We are honest, trustworthy and straightforward. We give plain-spoken, frank, accurate feedback and advice, and we never mislead or wilfully confuse.
```
dgmstuart: I feel like the original sentence used obfuscate in the sense of _hiding_ things from the client. 

The point is that while actively deceiving/misleading/confusing the client is obviously wrong, hiding things from the client (e.g. not telling them the full detail of a situation) might be considered pragmatic, so this sentence reinforces it as something we decide to not do.

RobjS: Ok, that makes sense. How about something like `we never mislead, wilfully confuse or obscure the issue at hand`?

dgmstuart: Do we need "wilfully confuse"? What's the motivation behind that?

Obscure works, but is it any plainer than obfuscate? ;P

alexdxw: how about 'complicate things'?

dgmstuart: @alexdxw is that relevant? We're talking about replacing 'obfuscate'. Complicating things feels like a separate issue to me.

dgmstuart: How about "we never mislead, or obscure the issue at hand?" - I feel like that preserves the original intention.

RobjS: To me, part of the meaning of `obfuscate` is clouding the issue by bombarding someone with information - to blind with science, in other words. So that's what I was trying to capture with `wilfully confuse`.

dgmstuart: That's not how I understand the meaning of that word, but regardless, I feel like we should take our steer from what the _intent_ of this paragraph is, rather than what the previously used word means. 

For me, the point is that we don't tell clients that things are OK when they're not, and we don't hide things from them

RobjS: I think `obscure the issue` probably covers both bases, anyway. So I'm happy with `we never mislead or obscure the issue at hand`.

alexdxw: I meant replace 'and we never mislead or obfuscate.' with 'and we never mislead or complicate things.'

tahb:
@dgmstuart this is waiting for you to review the changes your requested

PR #82, opened 2017-02-07
(feature) Playbook now contains long-term and repeated sick policies

harry-m:


harry-m:
Vita said:  I think '7 calendar days' should be clarified because I suspect that's talking about 7 working days in a row rather than 7 calendar days over a period of time?

harry-m:
No, this can be merged.

PR #85, opened 2017-02-16
When deploying, pull before merging into master

dgmstuart:
The current instructions say to fetch master but not to pull it before merging.

Fetching updates the dev's view of which commit the remote master is pointing at, but doesn't update their local master branch, so that may be behind.

If the dev's local master branch is behind origin/master then following these instructions to the letter will result in the incorrect merge happening. I _think_ this should lead to a failure when pushing (remote and local master will now both have commits which the other doesn't have), and it is recoverable, but will probably lead to a messy and confusing commit history.

### DIFF
--- a/guides/development-workflow.md
+++ b/guides/development-workflow.md
@@ -190,6 +190,7 @@ To deploy:
   ```
   git fetch
   git checkout master
+  git pull
   git merge origin/develop
   git push
   ```


guides/development-workflow.md
```diff
@@ -190,6 +190,7 @@ To deploy:
   ```
   git fetch
   git checkout master
+  git pull
```
tahb: Pixel Tom pointed out that git pull would create merge commits when I originally included `git pull` in this list:

> Incorporates changes from a remote repository into the current branch. In its default mode, git pull is shorthand for git fetch followed by git merge FETCH_HEAD

https://git-scm.com/docs/git-pull

What do you think?

dgmstuart: @tahb is that only because you have `--no-ff` on by default? 

Do you agree that we only ever want to merge develop into the latest version of master (the one on origin) - if so how can we do that without pulling?

dgmstuart: @tahb also, how do you normally get changes from remote branches? - e.g. to pull down a branch for code review.

dgmstuart: @tahb we clearly need to talk about this further - as a whole or partial team. One thing I strongly believe is that any process requiring everyone to have their local tooling configured a certain way is likely to fail.

dgmstuart: @holizz can you shed any light?

tahb: @dgmstuart My Tower merges are configured to use `--no-ff` and my CLI is default.

I believe we do only want to merge into the latest version of master.

Normally to get others work I fetch then checkout the remote. When somebody pushes more to that branch I will fetch, and pull.

I agree, we shouldn't have to rely on on personal configurations being made.

dgmstuart: @tahb 👍 

What happens when you have a local version of the branch and then checkout the remote? Do you delete your local branch in order to replace it with the remote version?

dgmstuart: :%s/@holizz/@tomdxw/g

tomdxw: @dgmstuart `git pull` causes merge commits in some circumstances. When `--no-ff` is set it always creates merge commits when master != origin/master.

But when that option isn't set, it probably won't happen very often, if everybody is doing everything else correctly. But I would still advocate using `pull --ff-only`. This would avoid situations where somebody's committed something to master they didn't meant to be on the master branch, or where something has otherwise gone wrong.

dgmstuart: @tomdxw I would be fine putting that into the dev guide:

```
git checkout master
git pull --ff-only
```

@tahb what do you think?

I can't imagine why anyone would _pull_ with `--no-ff`. That would be saying "I want a record in the main history of the project that I pulled this branch down onto my machine".

An alternative approach which bypasses some of the concerns would be:

```
git fetch origin
git checkout master
git reset --hard origin/master
```

That says "ditch whatever changes I have to local master: just point it at whatever origin is pointing at"

But in the (rare, probably not worth accounting for) case that someone has accidentally committed something to local master, this would blow those changes away and it would be reflog spelunking time. A plus of `pull --ff-only` is that it _errors_ when something has gone wrong.

tahb: The aim of documenting a single golden command that will guarantee the state of our code in every scenario is a worthy one but it feels like we also have to be realistic about it's application. 

I feel that everyone on the team should firstly understand and agree with our intentions, otherwise everyone will carry on with their current methods (which are fine as there are no deployment issues I'm aware of). I for one would continue to trust that anything on master was ready to go (as everything goes via PR) and fetch and pull as normal as that works for me. For these reasons I'm not particularly keen on adding either right now.

dgmstuart: @tahb Ok, do you agree that the current instructions need to be changed though? Checking out local master without pulling, and merging into that will always give the wrong result unless your local master matches remote master. For me that means that the current instruction is worse than no instruction at all!

dgmstuart: @tahb I don't want this to get stale...

tahb: Yeah I do agree we should put the pull in 👍

PR #86, opened 2017-03-17
(feature): The approaches we follow and tools we use to ship good code, often.

tahb:
* There was nothing in the playbook about TDD, CI or CD, all of which we have and use. This PR adds describes the foundations of what processes and tools we use in development.

### DIFF
--- a/playbook.md
+++ b/playbook.md
@@ -136,6 +136,26 @@ At a high level, the purpose of each phase is:
 
 We don't always do all these stages for every client. Often, we'll only be involved in a couple. But we can help with all or any of them, depending on our clients' needs.
 
+### Development
+At dxw, we follow several practices to help us regularly deliver code that is high quality, safe and maintainable, whilst preventing as many [regressions](https://en.wikipedia.org/wiki/Software_regression) in our software.
+
+#### Test-driven development
+When we pickup a new story developers start by writing tests that describe what the user needs to be able to do for the feature to be complete.
+
+These tests will initially all fail and as we write code they will begin to pass. After each change we make we run the tests again and if they go green, we can are done. This ultimately results in us writing less code which more features are shipped and the codebase is simpler.
+
+#### Automated testing
+We have tools set up to continuously run tests on active branches when there are changes and report back to the team via GitHub or Slack with what is passing and failing. This helps us stay on top of bugs that creep in unexpectedly when for example, two features have overlapped.
+
+#### Code reviews
+To keep the quality high and our bugs low, each new feature needs to be [looked at](../guides/development-workflow#code-review.md) by a peer. They will be looking to feedback on specific characteristics of the code such as bugs, security concerns, optimisations, readability and if the feature has complete [acceptance criteria](https://en.wikipedia.org/wiki/Acceptance_test%E2%80%93driven_development#Acceptance_criteria_and_tests).
+
+#### Automated deployment pipelines
+To keep our developers focused on building features that delivery value to users we invest time in automating the rest of our process.
+
+As part of the approach of [continuous delivery](https://en.wikipedia.org/wiki/Continuous_delivery) we want to safely [deploy](../guides/development-workflow##deploying.md) to different environments such as test, stage and production as often as possible.
+
+We use automation tools such as Jenkins, Chef and Codeship to manage this. Having machines take care of this regular task removes the risk of human error, ensures all tests are passing and there was at least one approved code review, without exception.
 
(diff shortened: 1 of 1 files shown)


playbook.md
```diff
@@ -136,6 +136,26 @@ At a high level, the purpose of each phase is:
 
 We don't always do all these stages for every client. Often, we'll only be involved in a couple. But we can help with all or any of them, depending on our clients' needs.
 
+### Development
+At dxw, we follow several practices to help us regularly deliver code that is high quality, safe and maintainable, whilst preventing as many [regressions](https://en.wikipedia.org/wiki/Software_regression) in our software.
+
+#### Test-driven development
+When we pickup a new story developers start by writing tests that describe what the user needs to be able to do for the feature to be complete.
+
```
RobjS: Hmm, this doesn't really describe how we do TDD on the PHP side I don't think. Or, to be honest, what I understood by TDD according to Uncle Bob's 3 rules: http://butunclebob.com/ArticleS.UncleBob.TheThreeRulesOfTdd.

playbook.md
```diff
@@ -136,6 +136,26 @@ At a high level, the purpose of each phase is:
...
+These tests will initially all fail and as we write code they will begin to pass. After each change we make we run the tests again and if they go green, we can are done. This ultimately results in us writing less code which more features are shipped and the codebase is simpler.
+
+#### Automated testing
+We have tools set up to continuously run tests on active branches when there are changes and report back to the team via GitHub or Slack with what is passing and failing. This helps us stay on top of bugs that creep in unexpectedly when for example, two features have overlapped.
+
+#### Code reviews
+To keep the quality high and our bugs low, each new feature needs to be [looked at](../guides/development-workflow#code-review.md) by a peer. They will be looking to feedback on specific characteristics of the code such as bugs, security concerns, optimisations, readability and if the feature has complete [acceptance criteria](https://en.wikipedia.org/wiki/Acceptance_test%E2%80%93driven_development#Acceptance_criteria_and_tests).
+
```
RobjS: Documenting what we mean by "Code review" is a big 👍 from me. @techbelly has some good ideas on this I think.

playbook.md
tahb: Thanks for this feedback, you're right I've started talking about testing behaviour and haven't presented my point properly. It also highlights a difference in approach that's worth a discussion, I think.

I will just say that I do agree with Uncle Bob's rules on TDD. 

At dxw we talk a lot about TDD but never BDD. The team on MyTVH was coached some time ago and were introduced to the additional idea of writing feature tests before any unit tests. It helped with immediate issues around focus on delivery, pain of context switching and reducing time spent down implementation rabbit holes.  

It became clear that without an emphasis on behaviour you risk spending days of writing code that doesn't actually provide the correct feature to users.

There are likely much better descriptions around but here's my take on what translating user needs directly into broad feature tests first helps us achieve and why I want to share it in the playbook:

1. they provide complete test coverage from the users perspective, if the test passes then we know that the user can click through a journey and fulfil their needs - we know when the card is done (short of any refactoring) and avoid implementation rabbit holes
2. they document all the features of a product, you can scan the feature test directory and at a glance understand what the service _does_
3. as they are broad they ensure that we always have a reliable test harness to refactor against should we want to change any of the implementation 
4. it steers us into writing only the units that are actually required

Renaming this section `Testing` may also serve as a better title.

I will hold off on editing this further until we've discussed further :)

playbook.md
```diff
@@ -136,6 +136,26 @@ At a high level, the purpose of each phase is:
 
 We don't always do all these stages for every client. Often, we'll only be involved in a couple. But we can help with all or any of them, depending on our clients' needs.
 
+### Development
+At dxw, we follow several practices to help us regularly deliver code that is high quality, safe and maintainable, whilst preventing as many [regressions](https://en.wikipedia.org/wiki/Software_regression) in our software.
+
+#### Test-driven development
+When we pickup a new story developers start by writing tests that describe what the user needs to be able to do for the feature to be complete.
+
+These tests will initially all fail and as we write code they will begin to pass. After each change we make we run the tests again and if they go green, we can are done. This ultimately results in us writing less code which more features are shipped and the codebase is simpler.
```
deberny: do you mean "we are done"?

CristinaRO:
This PR hasn't had any discussion in over six months, so we've decided to close it (Dev forum Oct 19 2018).

You can always reopen it or open a new one, which would also get rid of the conflict with the current master branch.

PR #90, opened 2017-09-28
Remove semicolons from commit labels

dgmstuart:
originally the intention was to use eg. `(feature):` but we decided long ago to use `(feature)` instead.


dgmstuart:
Huh - odd. I guess it's been that long since Michael or I worked on projects with you? It saves a character, which helps with staying under the 50 character limit, and it doesn't add anything, since the brackets are already marking the commit type

PR #93, opened 2017-11-13
naming-structure

vanessa-w:
Adding the dxw document naming convention

### DIFF
--- a/playbook.md
+++ b/playbook.md
@@ -899,6 +899,11 @@ We use Google Apps for Work to manage calendars, write and share documents. Ther
 
 When we write new things, we try to save them in a sensible folder within the existing structure.
 
+The dxw naming structure for documents is:
+
+Year-Month-Day    Doc type   Title
+E.g. 2017-01-01 Order Summary blog.gov.uk
+
 
 ### Protective marking scheme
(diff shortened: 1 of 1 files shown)


playbook.md
```diff
@@ -899,6 +899,11 @@ We use Google Apps for Work to manage calendars, write and share documents. Ther
 
 When we write new things, we try to save them in a sensible folder within the existing structure.
 
+The dxw naming structure for documents is:
+
+Year-Month-Day    Doc type   Title
+E.g. 2017-01-01 Order Summary blog.gov.uk
```
dgmstuart: @vanessa-w What should this date correspond to? The date it was originally created? The date it was sent? (if external)

vanessa-w:
The naming structure is used in our internal and client documentation. 
It's printed and up in the kitchen and new starters are made aware of it.  
This was to avoid the frequent use of titles such as "Contract" or "Order Summary" which was making filing frustrating.
The individual will choose the most appropriate date - be it the date of a contract or the date they created a document. We were aiming for the guidance to be as light touch as possible, given the subject.

dgmstuart:
@vanessa-w Thank you - that's a very clear explanation. If we could add some more words to this section to reflect that this applies to internal and client docs then we can get this merged in.

dgmstuart:
Superseded by https://github.com/dxw/playbook/pull/103

PR #94, opened 2017-12-14
Add new analytics guide

RobjS:


duncanjbrown:
This is all good, and I agree with the structure. I have a couple of things to add, and I'm sure more will come up. Shall we build on this branch and then get an external reviewer to merge?

RobjS:
Yep, I think that makes sense.

duncanjbrown:
Rebased and added a couple of commits

guides/web-analytics.md
```diff
@@ -0,0 +1,41 @@
+# Web analytics
+
+We use Google Analytics on most of our sites. It's free ([within limits](https://support.google.com/analytics/answer/1070983?hl=en)), it's easy to implement and there aren't many things it can't track.
+
+The login details for access are kept in 1Password, search for "Google Analytics".
+
+## Implementing
+
+We add Google Analytics via [gtag.js](https://developers.google.com/analytics/devguides/collection/gtagjs/). If you're using [whippet-theme](https://github.com/dxw/whippet-theme), the code is already included and you just have to enter the correct tracking ID.
```
tahb: This is exactly what I need. I would have otherwise I assumed this would be automatically available after plugging it in :+1

guides/web-analytics.md
```diff
@@ -0,0 +1,41 @@
...
+
+### Event tracking
```
tahb: This is exactly what I need. I would have otherwise I assumed this would be automatically available after plugging it in :+1:

guides/web-analytics.md
```diff
@@ -0,0 +1,41 @@
...
+
+If there are specific events you want to track that Google Analytics will not track by default (e.g. clicks on links to external sites, file downloads), you will need to implement a custom event, as per [Google's instructions](https://developers.google.com/analytics/devguides/collection/gtagjs/sending-data).
+
+It's important to note the section on [handling timeouts](https://developers.google.com/analytics/devguides/collection/gtagjs/sending-data#handle_timeouts). If you don't do that, the links you are tracking won't work if GA doesn't load (e.g. if the user has a browser extension that blocks trackers).
```
tahb: Is this only the case if you follow their guide to the letter and prevent default whenever we want to track an event? What is the benefit of interrupting the normal flow in order to send an event at all? Can we collect more information about the event by having it in a callback? 

As I see it if GA fails to load we won't get notified either way so the following seems like a reliable place to start, but I'm sure there's more capability here I'm missing:

```
  $ ->
    $('.form').click() ->
      gtag('event', 'form_complete', {});
```

guides/web-analytics.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Configuring
+
+There are a few basic steps it's worth following for any site you've set up in Google Analytics:
+
+* Set up an unfiltered view that records all traffic, including spam. You won't use this for most analysis, put it provides a control you can compare against, and a backup in case you get the filtering wrong on other views.
```
tahb: How do I create views?

tahb: I had to go to `admin` and on the right of the 3 columns select the default 'All Web Site Data' and 'create new view'.

guides/web-analytics.md
```diff
@@ -0,0 +1,41 @@
...
+* Set up a "main" view that has Google's "Bot filtering" activated (found under Admin > View Settings). This should automatically remove the majority of spam traffic.
+* You may also want to filter out internal traffic on the main view (e.g. from within the dxw office, or staging/localhost versions of the site). You can do this by IP and hostname.
```
tahb: Took me a few moments to realise I needed to switch to the Google Filters instead of try and filter from the view settings. I found the video here informative: https://support.google.com/analytics/answer/1033162?hl=en

CristinaRO:
@RobjS @tahb It looks like this PR only needs one more push to get it approved and merged! Could either of you confirm what kind of review does it still need (technical, spelling and grammar, any other kind)?

isratc:
@RobjS 
Heey looks awesome,

Tiny feedback:
Whippet Theme link broken.
Add link to the Whippet Theme file that includes the GA code: https://github.com/dxw/whippet-theme-template/blob/master/app/Theme/Analytics.php

Under Configuration
Modify the text slightly so that the user is aware that by default a new view is unfiltered
And on point three:  
Perhaps in the the example given, when you mentioned IP and hostname < specify further that staging/local host goes under hostname?

Issue #41, opened 2016-08-24
Publish the Playbook using github.io on a custom domain


harry-m:
We need to write a script that updates this playbook's HTML version and point playbook.dxw.com at it.

See #10 for more details.

harry-m:
This is now working: http://playbook.dxw.com/

Remaining issues:

* Deploy script needs to pull before pushing
* Add instructions to readme

dgmstuart:
@harry-m @tomdxw could one of you please check the attached pr (#96)?

PR #96, opened 2018-01-05
Fix deployment script, add deployment to readme

dgmstuart:


dgmstuart:
Fixes https://github.com/dxw/playbook/issues/41

build/deploy
```diff
@@ -3,6 +3,7 @@
 git checkout master && \
   git pull origin master && \
   git checkout gh-pages && \
+  git pull origin gh-pages && \
   git merge master && \
```
tomdxw: The merge command should have a `--no-edit` on it otherwise it spawns the user's editor when it's a non-ff merge.

build/deploy
```diff
@@ -3,6 +3,7 @@
 git checkout master && \
   git pull origin master && \
   git checkout gh-pages && \
+  git pull origin gh-pages && \
```
tomdxw: Are we sure we want to potentially create merges with these two `pull` commands? How about `pull --ff-only` for both of these?

README.md
```diff
@@ -44,3 +44,11 @@ This is a public document, because it's good to be open. But some things are pri
 
 ### Headings
 Use second level headings for main sections, third for subsections and fourth level ones if you need to. Don't use fifth or first level headings. The table of contents in the HTML version is automatically generated, and it assumes this structure.
+
+## Deployment
+
+To compile the playbook and deploy to github pages run:
+
+```
+build/deploy
```
tomdxw: The script assumes that you're in the `build` directory so this doesn't work.

build/deploy
```diff
@@ -3,6 +3,7 @@
 git checkout master && \
```
tomdxw: This script could be improved by adding this just below the shebang line:

```
set -e
```

This will make it so that any non-zero return code causes the script to cease execution, which means you can remove all the `&& \`.

build/deploy
dgmstuart: For `gh-pages` maybe this should be `git reset --hard origin/gh-pages`, since nobody should ever be committing to `gh-pages` except to deploy, so nobody should care about local changes?

Regarding master, depends who's going to be deploying I guess? Since the fix would be to rebase master. Presumably the only way you could fail with `pull --ff-only` is if you'd committed locally and not pushed those changes?

README.md
dgmstuart: Amended the original commit

build/deploy
dgmstuart: I think that's another point in favour of doing a reset instead of a pull

dgmstuart:
@tomdxw Pushed some new commits which hopefully address all your concerns.

tomdxw:
Looks good, and works.

PR #110, opened 2018-03-20
Be consistent in our use of PipelineDeals

rjw1:
and follow their branding


rjw1:
@CristinaRO Updated to include the one I missed.

PR #111, opened 2018-04-04
Link to HTML version instead of Markdown in Github

harry-m:
This removes the link in the readme to the Github version of the playbook and links to the nice one instead.


rjw1:
The link in the repo settings could be changed to using https://playbook.dxw.com as well.

rjw1:
My power got elevated so I fixed that link.

PR #127, opened 2018-06-29
Fix Markdown linting errors

robbiepaul:
This PR contains fixes to get the linting in #126 passing.

To make reviewing the changes a little easier use the [side by side view](https://github.com/dxw/playbook/pull/127/files?utf8=%E2%9C%93&diff=split&w=1) or view each commit separately.

Trello: https://trello.com/c/r2D2d5xh/3-add-linters-for-the-playbook


contributing.md
```diff
@@ -1,70 +1,80 @@
 # Contributing to the Playbook
 
 ## What you will need
+
 1. A GitHub account
-2. To be invited to the [dxw organisation](https://github.com/dxw)
-3. To be invited to the [staff team](https://github.com/orgs/dxw/teams/staff)
+1. To be invited to the [dxw organisation](https://github.com/dxw)
```
evilstreak: This seems weird – I know it's valid Markdown, but it stops it scanning nicely in plain text format. Is it down to a particular rule we might be able to disable? What do others think?

rjw1: The argument for doing it as all `1.` is that adding and removing and reordering items is less painful. the rendered version will do the right thing.
I personally think that actually all ordered lists should just be unordered ones since do the numbers actually add anything.

evilstreak: Seems reasonable to me. Definitely agree on using unordered lists in most places.

rjw1:
:shipit: :ship:ed

PR #135, opened 2018-08-02
Update Tech support info

olivia-work:
Correcting syntax

### DIFF
--- a/playbook.md
+++ b/playbook.md
@@ -1834,9 +1834,10 @@ technology decisions and build better services
 
 ### Out of office email message
 
-If you require an out of office message on your email account then you will
-need to send your request to support@dxw.com with the start and return date
-for this to be added to our mail server.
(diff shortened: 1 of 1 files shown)


playbook.md
```diff
@@ -1832,6 +1832,18 @@ Mobile - if you’re comfortable to share
 www.dxw.com :: 0345 2577520 :: @dxw helping public sector teams to make good
 technology decisions and build better services
 
+### Out of office email message
+
+If you require an out of office message on your email account then you will
+need to send your request to support@dxw.com with the start and return date
+for this to be added to our mail server.
```
vanessa-w: I would also suggest adding "and the contact details for another member of the team for any urgent queries" after "return date".

olivia-work: I've amended to include your suggested wording, please review.

PR #140, opened 2018-10-10
Ask people to record holiday in calendar

evilstreak:
Add clearer reasons for the different steps of the holiday process and add a step about updating your own calendar.

Understanding why the process is as it is helps people see the value in following it, and makes it easier to challenge or change parts of it.

### DIFF
--- a/playbook.md
+++ b/playbook.md
@@ -1447,6 +1447,9 @@ If you want to take time off:
   the impact it will have on the team and the project
 * **Request the holiday through BreatheHR**: we use BreatheHR to track who's
   off when so we can plan for it
+* **Add your holiday to your calendar**: if someone is trying to find out where
+  you are or when you're available the first thing they'll check is your
+  calendar
 
 We try very hard to make sure that all holiday requests are granted,
 irrespective of when they are made. But it's much easier for delivery managers
--- a/playbook.md
+++ b/playbook.md
@@ -1441,11 +1441,12 @@ For any other salary queries, please speak to your head-of or the Managing Direc
 
 ### Holiday
 
-We use BreatheHR to manage holiday requests.
+If you want to take time off:
 
-If you want to take holiday, first discuss the dates you plan to be away with
-your team so that they know when you're planning to go. Then request the holiday
-through BreatheHR.
+* **Discuss the dates with your team**: this gives them a chance to think about
+  the impact it will have on the team and the project
+* **Request the holiday through BreatheHR**: we use BreatheHR to track who's
+  off when so we can plan for it
 
 We try very hard to make sure that all holiday requests are granted,
 irrespective of when they are made. But it's much easier for delivery managers


playbook.md
```diff
@@ -1441,11 +1441,15 @@ For any other salary queries, please speak to your head-of or the Managing Direc
...
-If you want to take holiday, first discuss the dates you plan to be away with
-your team so that they know when you're planning to go. Then request the holiday
-through BreatheHR.
+If you want to take time off:
+
+* **Discuss the dates with your team**: this gives them a chance to think about
+  the impact it will have on the team and the project
+* **Request the holiday through BreatheHR**: we use BreatheHR to track who's
+  off when so we can plan for it
+* **Add your holiday to your calendar**: if someone is trying to find out where
```
CristinaRO: This is one way to do it! However, a shared calendar populated from BreatheHR already exists.

I suggest instead of asking people to duplicate admin work we share the knowledge about the common calendar - currently Robbie found it in one Slack message from August.

We could put that in the playbook, or if we don't want the calendar URL to be public, pin it to #general in Slack.

rjw1: I'm not entirely sure people check calendars anyway.
The one advantage of adding it to your own is that people can see it when doing meeting invites. You can copy across from calendars to make it slightly easier.
Personally I'm just glad I no longer have to spend all my time playing find the time everyone can make a meeting game.

evilstreak: I'll write this up into the commit message, but in the meantime (and to help me brain-dump it?) the rambling version is:

I know about the shared calendar already, and it doesn't solve the problem I have.

I, unfortunately, spend rather more time than I would like trying to schedule my time for the week. Normally that scheduling of time involves at least one other person. To check when someone is available my instinct is to look at their calendar. Almost all the time that's good enough: some people do put time off into their own calendar, and even for those who don't it'll still be correct for about 90% of the year. When it goes wrong, it sometimes only gets caught at the last minute.

For concrete examples:

- I arranged a session to talk about dxw-digest with the three devs who've worked on it. The day I'd scheduled it for two of them were off, but as they're both normally remote I didn't realise until close to the time of the session, meaning a last minute change of plans for me and the third dev who was around.
- In August I had to rearrange lots of 1:1s to work around my own holiday, some client workshops and some away days. Being able to see time off recorded in each persons calendar, with the correct colour-coding, made juggling all of those a lot less painful for me.

Consider putting the event into your calendar as a cache to improve read performance. Your calendar data is likely to be read hundreds of times over the year by different people, but you'll only book holiday a handful of times. The cost to duplicate the admin work is lower than the cost of a) asking people to check multiple calendars every time they're scheduling something, or b) dealing with the errors when they don't check multiple calendars and make a mistake.

Does any of that rambling make sense? Does any of it seem horribly flawed?

CristinaRO:
The explanation makes sense. _If_ you have the chance to amend the commit message that would be great, otherwise feel free to merge.

I'd still like to have the shared calendar pinned somewhere where we (dxw internal) can easily find it, but the Playbook doesn't seem the best place for it.

PR #145, opened 2018-10-17
Correct email signature layout

vanessa-w:
Correct email signature layout


vanessa-w:
Further changes to this section are needed so a new pull request has been created.

PR #151, opened 2018-11-02
Feature/add prose linting

RobjS:
This uses https://github.com/errata-ai/vale to add prose-linting to the playbook.

It includes custom rules to cover the guidelines outlined in the playbook itself (https://github.com/dxw/playbook/blob/master/guides/style-guide.md) including the list of GOV.UK words to avoid.

There's also some basic checks for offensive terms.

Most of these will trigger warnings rather than errors, to prevent blocking deployment where the usage is legitimate, but a few (e.g. the offensive terms, not using the correct case for `dxw` or terms like `WordPress` & `GovPress`) will cause errors in CI until corrected.

### DIFF
--- a/styles/govuk/words-to-avoid.yml
+++ b/styles/govuk/words-to-avoid.yml
@@ -10,18 +10,18 @@ swap:
   combating: Avoid using 'combating'
   countering: Avoid using 'countering'
   deliver: >
-    Avoid using 'deliver'. Pizzas, post and services are delievered; not abstract concepts like improvements or priorities
+    Avoid using 'deliver'. Pizzas, post and services are delivered; not abstract concepts like improvements or priorities
   deploy: Avoid using 'deploy', unless it's military or software
   dialogue: Avoid using 'dialogue', unless we speak to people 
   disincentivise: Avoid using 'disincentivise'
   empower: Avoid using 'empower'
-  facilitiate: Avoid using 'facilitate', say something specific about how you're helping
+  facilitate: Avoid using 'facilitate', say something specific about how you're helping
   focusing: Avoid using 'focusing'
   foster: Avoid using 'foster', unless it's children 
   impact: Do not use 'impact' as a synonym for have an effect on, or influence 
   initiate: Avoid using 'initiate'
   key: Avoid using 'key', unless it unlocks something
-  land: Only use as 'land' as a verb if you're talking about aircraft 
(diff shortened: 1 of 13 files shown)


styles/govuk/words-to-avoid.yml
```diff
@@ -0,0 +1,38 @@
...
+level: warning
+ignorecase: true
+swap:
+  agenda: Avoid using 'agenda', unless it's for a meeting 
+  advancing: Avoid using 'advancing'
+  collaborate: Use 'working with' instead of 'collaborate'
+  combating: Avoid using 'combating'
+  countering: Avoid using 'countering'
+  deliver: >
+    Avoid using 'deliver'. Pizzas, post and services are delievered; not abstract concepts like improvements or priorities
```
CristinaRO: Typo: 'delievered'

Concept: isn't it a bit hypocritical, if some of our people are... "Delivery Manager"s? 🙈

styles/govuk/words-to-avoid.yml
```diff
@@ -0,0 +1,38 @@
...
+  deploy: Avoid using 'deploy', unless it's military or software
+  dialogue: Avoid using 'dialogue', unless we speak to people 
+  disincentivise: Avoid using 'disincentivise'
+  empower: Avoid using 'empower'
+  facilitiate: Avoid using 'facilitate', say something specific about how you're helping
```
CristinaRO: Typo: 'facilitiate'

styles/govuk/words-to-avoid.yml
```diff
@@ -0,0 +1,38 @@
...
+  focusing: Avoid using 'focusing'
+  foster: Avoid using 'foster', unless it's children 
+  impact: Do not use 'impact' as a synonym for have an effect on, or influence 
+  initiate: Avoid using 'initiate'
+  key: Avoid using 'key', unless it unlocks something
+  land: Only use as 'land' as a verb if you're talking about aircraft
```
CristinaRO: Typo: extra 'as' before 'land'

styles/govuk/words-to-avoid.yml
RobjS: I assumed our Delivery Managers fall under the "services are delivered" exception, but it does seem ambiguous.

RobjS:
Thanks for spotting the typos @CristinaRO! Should be fixed now. I've also updated the README with instructions on running the prose linter.

PR #172, opened 2019-01-16
Add guide for support and on call

evilstreak:
This is a first version of some documentation around our support and on call processes and what it means for the people doing the work.

Huge thanks to @Ellayanor for pulling these words together from various places and getting them into shape 🙏.

Due to the way the headings get turned into a nav sidebar the long question form isn't ideal, but I couldn't bring myself to strip the personality from it to use three word titles. I think it's probably fine:

screenshot 2019-01-16 at 18 19 49

This should be considered a first draft. I expect it to be changed a fair bit in the next month as we get ready to do the first line support rota for real. Given that, I'm mostly interested in hearing about whether there's anything in here you think is wrong or misleading, or shouldn't be made public. Things that are missing or that could be improved should be picked up in another pull request.

I'd particularly like @rjw1, @leedxw and @claredxw to look at this.

### DIFF
--- a/guides/support-and-on-call.md
+++ b/guides/support-and-on-call.md
@@ -1,0 +1,154 @@
+# Support and on call
+
+We have service definitions for [GovPress hosting], [container hosting] and
+secure WordPress hosting for our [blogging platform] and [campaign multisites]
+through G-Cloud.
+
+We have agreed service levels for each client we provide support for. Our
+standard terms, including prioritisation definitions and response times, are
+detailed in our [Service Level Agreement].
+
+With our default support plan we will manage all code, maintenance, updates and
+uptime monitoring and provide security event monitoring intended to detect and
+prevent attacks. Users can also use our helpdesk service for help with any
+aspect of their site.
+
+Our Managed Plus plan is for clients with higher assurance requirements. It
+includes all the fully-managed features of our default support plan, plus a
+quarterly site review meeting, a named delivery manager and an enhanced SLA
+with faster response targets.
+
+We provide support for clients and users of these services in three ways:
+
+* Clients and users of these services can raise a **ticket** with us about
+  usability issues, errors or faults. Sometimes what they’re asking for is a
+  larger feature request, which needs separate handling.
+* We do proactive **maintenance** for our client’s sites and services, for
+  example by applying security updates.
+* We have monitoring systems for our services that continually check whether
+  they are operating correctly, and whether there are any warning signs of
+  problems. When problems are detected they raise an **alert** for the team to
+  investigate and resolve.
+
+That support is handled by the team in two ways:
+
+* We provide in-hours (10.00–18.00) support for all our clients, resolving
+  tickets and alerts and doing maintenance. We call this **first line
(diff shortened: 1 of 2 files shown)


guides/support-and-on-call.md
```diff
@@ -0,0 +1,156 @@
...
+* We provide out-of-hours (24/7) support for some of our clients, responding
+  only to critical alerts and emergency tickets. We call this **on call**, and
+  it’s staffed by a developer or an ops engineer through an opt-in rota.
+
+## Being on first line support
+
+Support tickets come in from clients and users every day. These tickets could
+relate to any of the products we currently support, which are a mixture of
+rails apps and Wordpress implementations, spread across clients all over the
+private sector, including central government.
```
mec: public??

guides/support-and-on-call.md
```diff
@@ -0,0 +1,156 @@
...
+  support**, and it’s staffed by a developer and an ops engineer on rota.
+* We provide out-of-hours (24/7) support for some of our clients, responding
+  only to critical alerts and emergency tickets. We call this **on call**, and
+  it’s staffed by a developer or an ops engineer through an opt-in rota.
+
+## Being on first line support
+
+Support tickets come in from clients and users every day. These tickets could
+relate to any of the products we currently support, which are a mixture of
+rails apps and Wordpress implementations, spread across clients all over the
```
lozette: Capitalise Rails

guides/support-and-on-call.md
```diff
@@ -0,0 +1,156 @@
...
+
+While on support, your focus is on monitoring incoming alerts and tickets,
+responding to them or delegating to colleagues as appropriate.
+
+#### Alerts
+
+We use monitoring and alerts to notify you when an issue occurs, and provide
+[run-books] to help you respond to events and restore service for our clients.
+
+[run-books]: https://git.govpress.com/ops/docs
```
lozette: Should this be an actual link?

guides/support-and-on-call.md
RobjS: And "Wordpress" should be "WordPress"

guides/support-and-on-call.md
evilstreak: If you mean in terms of markdown syntax, this provides a link for the reference (`[run-books]`) in the paragraph above (and also another paragraph further down the page).

If you mean in terms of that URL doesn't resolve to anything for you, then we'll need to make sure you have the right access permissions for it.

guides/support-and-on-call.md
evilstreak: Thanks! Fixed

guides/support-and-on-call.md
evilstreak: Thanks both! Fixed

guides/support-and-on-call.md
lozette: Ah my mistake, I misunderstood the syntax!

evilstreak:
I've moved all the link references to the bottom. I'll merge when green. Thanks all!

PR #178, opened 2019-02-07
(Chore) Added delivery leads guide for arranging pen tests

olivia-work:


guides/delivery/delivery-lead-guide-pen-tests.md
```diff
@@ -0,0 +1,13 @@
+# Delivery leads guide for arranging pen tests
+
+When a penetration test is arranged by a client on a site which is hosted by
```
lozette: This sentence feels like a run-on sentence. Maybe add a couple of commas: 

`When a penetration test is arranged by a client on a site which is hosted by dxw digital, the delivery leads must find out the scope of the test and raise a ticket in Zendesk, as the Ops team will need to notify AWS as part of our contract with them.`

staceydxw: Yes that looks better

staceydxw:
Bit confused now about whether I can make the changes and merge or whether someone else needs to do it. Let me know if I can help :)

guides/delivery/delivery-lead-guide-pen-tests.md
CristinaRO: @olivia-work Given that this is your PR, would you mind one of us making the above changes, or do you want to make them yourself?

olivia-work:
This document is no longer required because AWS has announced that we no longer need to notify them when a pen test is taking place.

PR #184, opened 2019-02-10
Make principles more consistent

johnwaterworth:
Turn principles into lists. Add custom heading ids for linking


johnwaterworth:
I've tried a few things to fix the linting errors on the nested lists, with no success. The markdown follows the example in the Github documentation and the examples in the error descriptions.

lozette:
Good grief that was way more painful than it should have been!! Turns out there were mixtures of tabs and spaces in the list indentations and the linter didn't like it.

PR #192, opened 2019-02-18
(fix) Benefits - mention probation period

rkateren:
making clear benefits are available after passed probation period.

### DIFF
--- a/playbook.md
+++ b/playbook.md
@@ -2230,7 +2230,7 @@ Slack channel.
 
 ### Benefits
 
-Subject to successful probation.
+All perks and benefits are available to the dxw team after the successful passing of their probationary period.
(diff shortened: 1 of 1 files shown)


playbook.md
```diff
@@ -2230,6 +2230,8 @@ Slack channel.
 
 ### Benefits
 
+Subject to successful probation.
```
CristinaRO: Would it be clearer to make it into a full sentence, such as "Benefits start applying on the day" (or "after the day", I don't know which) "you pass your probation" ?

rkateren: Yes, Something like 'All perks and benefits are available to the dxw team after the successful passing of their probationary period.'

PR #210, opened 2019-02-28
Remove an unclear possibly prejudicial line from Screening criteria

erbridge:
Discussed with @vanessa-w.

When I was applying, this line made me very uncomfortable. It
doesn't give any guidance on what the requirements are. This is
probably one of the first sections a potential candidate might
read, and it might put of some of the more diverse applicants from
applying.

### DIFF
--- a/playbook.md
+++ b/playbook.md
@@ -1601,9 +1601,7 @@ qualities than about skills. Things to keep in mind:
 
 * Does this person seem likely to embody our [values](#values)?
 * Does this person have an interest in our [mission](#why-are-we-all-here)?
-* Does this person have experience working in [agile environments](#agile-mindset)
-* Would you be comfortable taking this person to a meeting with a client?
-  If not, why not?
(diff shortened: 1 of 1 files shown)


tahb:
I appreciate this wording is open to interpretation which is unhelpful and off putting. I'd like to see if we can try to improve this wording to provide clearer guidance before removing it. 

I believe the original intent here is to help remind us to also evaluate a candidate on their ability to:

- handle difficult situations, especially when there is conflict and disagreement
- articulate concisely, do they talk to the point rather than waffling off topic?
- present their work, are they able to communicate complicated work to folk who aren't expert in it? eg. show & tells

erbridge:
Is that appropriate in a screening?

Happy to reword instead of remove. @vanessa-w suggested removal, hence this PR.

tahb:
Quite right, sorry! I interpreted this as screening in general rather than the screening step, which it isn't appropriate for.

PR #211, opened 2019-02-28
Sort out the headings for the working arrangement section

erbridge:


playbook.md
```diff
@@ -1722,32 +1722,32 @@ there's an emergency that means we have to work during unsociable hours to solve
 the problem. Neither of these happen very often, but they are a normal part of
 life at dxw.
 
-### Flexible Working
+### Flexible working
 
 Our policy is informed by the way we work and the way we charge clients for our time.
 
-#### How we work
+### How we work
```
CristinaRO: I like the general tidying!

I think everything from "How we work" down to right before "Remuneration" was _meant_ to come under the broad heading of "Flexible working", which would make "Flexible working" the only h3 level before "Remuneration".

Otherwise, "Flexible working" is left as a h3 with only one line of text, which makes it look a bit not useful.

erbridge: Yeah, I thought about that, but the h5 formatting makes the headings blend in to the body (font size is smaller than the body text) and some of the sub-sections would then have h5s, so I opted for this in lieu of making them stand out better.

CristinaRO: 👍

PR #215, opened 2019-03-11
advisory council

farrahfa:
adding some words about the advisory council that dxw digital and cyber share


CristinaRO:
This looks great but still mysterious! Is there a reason we're not sharing their names? Is there an internal place where dxw employees can find out who are the members of the council? (Other than "Ask Dave" ;) )

farrahfa:
@CristinaRO we're just awaiting word from the members of the council we can go public with their names. Once we have that I will edit the words here. But great idea to have something internal, i'll put something together soon.

PR #220, opened 2019-04-08
Require that we note individual time use

leedxw:
This is a simple request that requires minimal changes to workflow.

We should also be investigating tool-based approaches, such as [Zendesk's own Time Tracking plugin](https://www.zendesk.com/apps/support/time-tracking/), assuming they don't introduce fresh annoyance.

### DIFF
--- a/playbook.md
+++ b/playbook.md
@@ -1107,8 +1107,8 @@ they probably won't see it.
 
 ##### Recording time spent on tickets
 
-It's important for us to know how much cumulaitive time is spent on support
-functions.  In order to mark a ticket as solved, a ticket the ticket owner needs
+It's important for us to know how much cumulative time is spent on support
+functions.  In order to mark a ticket as solved, the ticket owner needs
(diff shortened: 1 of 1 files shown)


playbook.md
```diff
@@ -1105,6 +1105,19 @@ approximation of the total time the team spent on the issue. Or, select
 When assigning a ticket to another agent, we do not set it to Pending, otherwise
 they probably won't see it.
 
+##### Recording time spent on tickets
+
+It's important for us to know how much cumulaitive time is spent on support
```
erbridge: Typo: `cumulative`.

playbook.md
```diff
@@ -1105,6 +1105,19 @@ approximation of the total time the team spent on the issue. Or, select
...
+functions.  In order to mark a ticket as solved, a ticket the ticket owner needs
```
erbridge: Extra `a ticket` here.

PR #221, opened 2019-04-12
Remove old linter

CristinaRO:
* Remove old linter from Travis build
* Configure remark similar to mdl
* Update readme with instructions to lint locally
* Remove old linting scripts and config


Gemfile
```diff
@@ -3,5 +3,3 @@
 source "https://rubygems.org"
```
erbridge: Since it doesn't look like Ruby's a dependency anymore, can we remove the Gemfile and lock entirely?

CristinaRO: Oh! Good point, thanks!

PR #230, opened 2019-06-12
Improve the Purpose and Principles

johnwaterworth:
Remove some repetition in the early sections, particularly in guidance about updating the Playbook.
Tighten up the Purpose statement to reflect a more mature Playbook
Make the Principles statements more consistent in how they are expressed
Start to make the Principles clearer by removing or moving detailed guidance that should live elsewhere

### DIFF
--- a/playbook.md
+++ b/playbook.md
@@ -2,28 +2,29 @@
 
 ## Purpose
 
-For most of dxw's life, we've been small enough that it's been easy for everyone
-to know what's going on and how to get things done, just by listening to what's
-going on in the office. The team is now big enough that we need a reference for
-the way we do things. A place to capture the changes we agree to make. Something
-canonical that tells us what the current "right way" to do things is.
+This Playbook is our reference for who we are and the way we do things. Something canonical that tells us what the current “right way” to do things is.
 
-We'll always iterate on this document, because we'll always explore better ways
-to get things done. But now, when we've found one, it'll be recorded here. And
-once it is, it'll be the way we do things until we find something better, and
-write it down.
+If you’re a current or potential client, this Playbook is also for you. To help you understand us and how we can work together.
 
-dxw now encompasses dxw digital and dxw cyber, many of the values and ways of working are the same across both businesses and so one document is used for both, with any differences being identified.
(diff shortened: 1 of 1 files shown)


playbook.md
```diff
@@ -119,31 +120,20 @@ of wisdom, but we don’t follow the industry of methodologies, training and
 certifications that has grown up around it. We think agile is something you
 learn to be, not something that you learn to do.
 
-Our teams frequently experiment with different tools and techniques so we can
-improve. This applies not only to development work, but to everything we do at
-dxw, from user research to business operations.
```
erbridge: :+1:

playbook.md
```diff
@@ -119,31 +120,20 @@ of wisdom, but we don’t follow the industry of methodologies, training and
...
+maintaining the trust that our clients give us, and it’s important not to let
 them down.
 
-When taking time off during client work, we discuss how it will impact the
-schedule with other team members.
-
-Sending communication outside of working hours may create an unintended sense of
-urgency with the recipients of your message. Try to avoid creating that urgency
-when possible. Unless actually urgent, you may ignore messages which you receive
-outside of working hours and handle them once you are back at work.
```
erbridge: How come we're removing this? I think it's an important callout to our culture of sustainability, and how it might impact team and client.

playbook.md
```diff
@@ -119,31 +120,20 @@ of wisdom, but we don’t follow the industry of methodologies, training and
...
+#### Balance openness and confidentiality
 
-#### Openness and confidentiality (dxw digital)
+### dxw digital
```
erbridge: Shouldn't these subheadings be a level deeper than the level above? i.e. `#####` rather than `###`.

johnwaterworth: Oops. Corrected the heading levels.

playbook.md
johnwaterworth: It is important stuff. But it's detailed guidance. We have a section later in the playbook about time and flexible working. We should integrate this material into those sections.

johnwaterworth: Moved the material into the Time section.

PR #233, opened 2019-07-01
Add new kick-off workshop guides to Guides menu

johnwaterworth:


build/partials/_navbar.md
```diff
@@ -13,6 +13,8 @@
   - [Plugin reviews](guides/plugin-reviews.md)
   - [Plugin updates](guides/plugin-updates.md)
   - [Setting up a new project](guides/delivery/setting-up-a-new-project.md)
+  - [Running a discovery kick-off workshop](guides/running-a-discovery-kick-off-workshop.md)
```
CristinaRO: Would you mind reordering the new items so the guides are still listed alphabetically, please?

johnwaterworth: Thanks Christina. I would argue that alphabetical is not a good order for the guides:
https://www.nngroup.com/articles/alphabetical-sorting-must-mostly-die/

CristinaRO: This can certainly be a topic for the playbook reorganisation project! But until we arrive at a more logical ordering, I'd argue in favour of consistency with what little order we do have.

johnwaterworth: OK. Alphabetical order for now.

PR #240, opened 2019-07-18
Added reporting our time to playbook

kathcooper3:
Description for reporting out time within 10,000ft

### DIFF
--- a/playbook.md
+++ b/playbook.md
@@ -1961,11 +1961,13 @@ We track our time at the same level of granularity that we bill it – for almos
 
 When reporting our time, we follow these guidelines:
 
-* If you worked on a named project (client or internal), record your time as that.
+* If what was scheduled matches reality, simply confirm this time.
+* If you worked on a named project (client or internal), that differs from the schedule, record your time as that.
 * If you worked on something else, leave the time blank or zero out what you were scheduled to work on.
+* If you were away on planned leave, confirm your time as "Out of office".
 * If you were off sick, record your time as "Sick".
 * If you had unplanned leave, record your time as "Out of office".
-* If you were away due to a regular day off, as part of your working pattern, leave the time blank.
+* If you were away due to a regular day off, as part of your working pattern, leave the time blank or zero out what you were scheduled to work on.
 
 If you’re not doing client work, then unless you’re working on one of a handful of named internal projects, you don’t need to record your time against any other piece of work – the assumption is you’re doing valuable work, whether it’s learning, helping with recruitment or improving our tooling and processes, and we don’t need itemised time tracking to justify that.
 
--- a/playbook.md
+++ b/playbook.md
@@ -1947,27 +1947,29 @@ incurring them.
 We manage expenses using Xero. For more information about how to do this, see
 the [guide on claiming expenses](/guides/claiming-expenses)
 
-## Reporting our time
+### Reporting our time
 
 Everyone who works on a client project at dxw digital is responsible for reporting the time they spend on projects. We usually work with our clients on a [time and materials basis](https://www.gov.uk/guidance/how-to-pay-for-digital-outcomes-and-specialists-services#time-and-materials), so it’s important we can accurately track and report how our time is spent on client projects.
 
-We use [10,000ft](https://app.10000ft.com/) to schedule our work, which helps us to forecast our capacity over future weeks and months in a reliable way. Each member of our delivery teams uses the tool to record the days they work on client projects.
+We use [10,000ft](https://app.10000ft.com/) to schedule our work, which helps us to forecast our capacity over future weeks and months in a reliable way. 
 
-We’ve set up the platform, so that it understands individual team members’ working patterns. This means the team don’t need to worry about entering in their non working days.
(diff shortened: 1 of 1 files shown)


playbook.md
```diff
@@ -1947,6 +1947,18 @@ incurring them.
...
+
+We use [10,000ft](https://app.10000ft.com/) to schedule our work, which helps us to forecast our capacity over future weeks and months in a reliable way. Each member of our delivery teams uses the tool to record the days they work on client projects.
+
+We track our time at the same level of granularity that we bill it – for almost everyone in the team, that’s half days or whole days.
+
+If you’re not doing client work, then unless you’re working on one of a handful of named internal projects you don’t need to record your time against anything – the assumption is you’re doing valuable work, whether it’s learning, helping with recruitment or improving our tooling and processes, and we don’t need itemised time tracking to justify that.
+
+Our delivery leads lend a helping hand when delivery team members aren’t sure how to record their time – they’ve been doing this for a long while and have a well developed sense for it.
```
erbridge: Since this came up for me, is it worth adding something about how we track (or not) Fridays, sickness, holidays, and other working pattern related non-working? For instance, I wasn't aware that I wasn't supposed to be using "Out of office" for my regular days off.

johnwaterworth: That's interesting, cause I thought we were supposed to be putting in our non-working days.

erbridge: I believe 10kft is now set up to understand part-time working schedules, so we only use "Out of office" for holiday. At least that's what I understood from talking to @evilstreak about it.

kathcooper3: @erbridge @johnwaterworth That's right - 10kft is set up to understand part-time working schedules. 

I'm going to write something about this, along with how we record sickness/annual leave, so this is clearer.

playbook.md
```diff
@@ -1953,9 +1953,21 @@ Everyone who works on a client project at dxw digital is responsible for reporti
...
+
 We track our time at the same level of granularity that we bill it – for almost everyone in the team, that’s half days or whole days.
 
-If you’re not doing client work, then unless you’re working on one of a handful of named internal projects you don’t need to record your time against anything – the assumption is you’re doing valuable work, whether it’s learning, helping with recruitment or improving our tooling and processes, and we don’t need itemised time tracking to justify that.
+If you’re not doing client work, then unless you’re working on one of a handful of named internal projects, you don’t need to record your time against any other piece of work – the assumption is you’re doing valuable work, whether it’s learning, helping with recruitment or improving our tooling and processes, and we don’t need itemised time tracking to justify that.
+
+We should however record planned leave and sickness. Our Business Operations team update the schedule to reflect any planned leave, which helps us to look ahead at where there might be skills gaps on client work- this is recorded as “Out of office”.
```
erbridge: Missing space before the `-`.

playbook.md
```diff
@@ -1953,9 +1953,21 @@ Everyone who works on a client project at dxw digital is responsible for reporti
...
+
+If we spend time away from the office sick or if we don’t end up working on a project that we had planned to, we need to reflect this too.
+
+When we record our time, we should;
```
erbridge: Did you mean to use a semicolon here?

kathcooper3: Ahh yes I did. Thanks!

playbook.md
johnwaterworth: I think 10000ft only understands some part time working patterns. For example, I work 4 days a week. My non-working day is normally Wednesday, but I do sometimes change day if needed. So I believe that it's useful for me to enter my non-working days to help with planning.

playbook.md
```diff
@@ -1947,6 +1947,32 @@ incurring them.
...
+We track our time at the same level of granularity that we bill it – for almost everyone in the team, that’s half days or whole days.
+
+#### What to do when reporting your time
+
+When reporting our time, we follow these guidelines:
+
+* If you worked on a named project (client or internal), record your time as that.
+* If you worked on something else, leave the time blank or zero out what you were scheduled to work on.
+* If you were off sick, record your time as "Sick".
+* If you had unplanned leave, record your time as "Out of office".
```
erbridge: I think it's helpful to also include planned leave here, since you're also confirming what was scheduled, and want to be leave scheduled "Out of office" as that.

playbook.md
```diff
@@ -1947,6 +1947,32 @@ incurring them.
...
+* If you were away due to a regular day off, as part of your working pattern, leave the time blank.
+
```
erbridge: Possibly worth noting that there is an action when the schedule matches reality, which is to confirm the schedule?

playbook.md
erbridge: In 10kft, if you're an admin, you can set per-week working patterns, so this perhaps this could be accomodated that way. For example mine has me manually set to 0 availability every other Monday. It's annoying that it's not possible to view what's set without modification access, though.

kathcooper3: I've been through 10,000ft and have added everyone's part time working patterns until December 2019, including those who work 3 days or 4 days per week and those who work different days on a fortnightly basis.

If someone changes their non working day, I think this should be down to the individual to record the day they actually worked on a client or named internal project, along with the day they weren't in the office as planned (as "Out of Office"), against the weekly reporting of their time - it's possible to record time on a day you shouldn't have been working on.

evilstreak:
Thanks @kathcooper3!

PR #241, opened 2019-07-22
(feature) Development principles

jcoglan:
This is the result of the drafting session that concluded on Friday 19
July 2019. We're now moving the draft to the playbook and all further
edits should be done here.

### DIFF
--- a/playbook.md
+++ b/playbook.md
@@ -753,6 +753,130 @@ Designers at dxw follow these principles:
 
    There is always room for improvement. Design can always be made better. There is a difference between ‘done’ and ‘finished’.
 
+### Development
+
+#### Development principles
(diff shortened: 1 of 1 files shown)


Floppy:
It's just occurred to me, we've talked about "development" principles, but this covers ops too. Should it be "technology" instead? @evilstreak?

evilstreak:
> It's just occurred to me, we've talked about "development" principles, but this covers ops too. Should it be "technology" instead? @evilstreak?

Nope, this should be development only. Whilst there is overlap between disciplines in the tech team, I think it's probably most flexible (at least to start with) for each discipline (design, development, ops engineering) to have their own.

PR #244, opened 2019-07-30
Update to hosting and supporting services section

kathcooper3:
Following a review of the support section of the playbook, information has been removed that either would be best placed as internal documentation, for instance within ops docs or the information is incorrect and needs to be re-worded to make it more accurate.

There is more to work on here - this update should reflect what should be removed, due to being out of date information or more internal focussed. Additions are a starting point to be iterated upon by others.

### DIFF
--- a/playbook.md
+++ b/playbook.md
@@ -971,30 +971,32 @@ to build an appropriate hosting platform for their service. This is fair
 enough, because they shouldn't have to. We built GovPress to meet this need,
 and it is now used by most of our clients.
 
+We build [websites](https://www.digitalmarketplace.service.gov.uk/g-cloud/services/930612236449495), [campaign](https://www.digitalmarketplace.service.gov.uk/g-cloud/services/691308552308120) and [blogging](https://www.digitalmarketplace.service.gov.uk/g-cloud/services/355674790119695) platforms hosted on GovPress, configured to our client’s branding requirements.
+
 You can read more about GovPress at
 [www.govpress.com](https://www.govpress.com).
 
-We're in the process of building a new container-based hosting platform which
-will be able to support services that aren't built in WordPress. Watch this
-space.
+### Tailored support
+
+We also offer support for services that aren't built in WordPress. We are able to host and support services with our [tailored support plan for cloud applications](https://www.digitalmarketplace.service.gov.uk/g-cloud/services/393702576050692), even if your service isn't hosted with us.
 
 ### Support helpdesk
 
 The dxw Technical Operations team build, manage and run GovPress, as well as
-provide first-line support.
+providing first-line support for our clients on tailored support plans.
 
 We use [Zendesk](https://dxw.zendesk.com) to manage support requests. Any
 incidents, requests for us to fix a problem or make a change to a site must
 come to us via Zendesk as a ticket. We use tickets to manage requests in order
 to:
 
-* Keep track of all the things we need to do, and what state they're in
+* Ensure oversight of all the things we need to do, and what state they're in, for us and our clients
 * Have a record of the changes we're asked to make
 * Ensure that we only accept change requests from people who are authorised
 * Generate data about how much staff time is spent on each client's issues.
 
 If a client asks us to do something in person, on the phone or via email, we
-politely ask them to visit the [dxw Zendesk](https://dxw.zendesk.com/) instead,
+ask them to visit the [dxw Zendesk](https://dxw.zendesk.com/),
 and submit a ticket there. This is so that:
 
 * We can keep track of the issue from report to solution, and ensure that it's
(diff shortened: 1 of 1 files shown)


playbook.md
```diff
@@ -971,189 +971,72 @@ to build an appropriate hosting platform for their service. This is fair
 enough, because they shouldn't have to. We built GovPress to meet this need,
 and it is now used by most of our clients.
 
+We build [websites](https://www.digitalmarketplace.service.gov.uk/g-cloud/services/930612236449495), [campaign](https://www.digitalmarketplace.service.gov.uk/g-cloud/services/691308552308120) and [blogging](https://www.digitalmarketplace.service.gov.uk/g-cloud/services/355674790119695) platforms hosted on GovPress, configured to our client’s branding requirements.
```
lexijax: I'd rephrase this a little - the service offer is for managed hosting and support rather than website build. Maybe something like 'We offer managed hosting and support for WordPress websites, as well as blogging and campaigns platforms, configured to our client's branding requirements, all on our GovPress hosting platform. 

Wonder if it's worth saying that we offer these services through the GCloud framework (as that's where the links are to)?

playbook.md
```diff
@@ -971,189 +971,72 @@ to build an appropriate hosting platform for their service. This is fair
...
+
 You can read more about GovPress at
 [www.govpress.com](https://www.govpress.com).
 
-We're in the process of building a new container-based hosting platform which
-will be able to support services that aren't built in WordPress. Watch this
-space.
+### Tailored support
+
+We also offer support for services that aren't built in WordPress. We are able to host and support services with our [tailored support plan for cloud applications](https://www.digitalmarketplace.service.gov.uk/g-cloud/services/393702576050692), even if your service isn't hosted with us.
```
lexijax: Wonder if there are actually two sections here - one for Dalmatian https://www.digitalmarketplace.service.gov.uk/g-cloud/services/754953588860938 
and one for the tailored support, as per your commit? 
Think it might make it a little clearer?

playbook.md
```diff
@@ -971,189 +971,72 @@ to build an appropriate hosting platform for their service. This is fair
...
-provide first-line support.
+providing first-line support for our clients on tailored support plans.
 
 We use [Zendesk](https://dxw.zendesk.com) to manage support requests. Any
 incidents, requests for us to fix a problem or make a change to a site must
 come to us via Zendesk as a ticket. We use tickets to manage requests in order
 to:
 
-* Keep track of all the things we need to do, and what state they're in
+* Ensure oversight of all the things we need to do, and what state they're in, for us and our clients
```
lexijax: Like the addition of 'for us and our clients'; prefer 'keep track' as it's plainer English.

playbook.md
```diff
@@ -971,189 +971,72 @@ to build an appropriate hosting platform for their service. This is fair
...
 is really important.
 
-#### Support sprints
-
-We work on tickets by having a developer on a support sprint. Developers are
-assigned to support for two weeks at a time and work on tickets for the
-duration of the sprint. If there is a ticket they need another developer to
-help with, this work is scheduled.
-
-#### Client experience (AKA: ticket principles)
```
lexijax: I like that we have these but agree that they need a bit of an update. Do we want to just trim them down for now and review with the team to sort out the detail, then we can update them properly later? I think the 'principles of how we do support' is a thing that should be in the playbook and I'm a little wary about removing it all now without having a commitment to replace them soon.

playbook.md
```diff
@@ -971,189 +971,72 @@ to build an appropriate hosting platform for their service. This is fair
...
-In this situation, it's only natural that we prefer to do other work first. But
-it is vital that we don't let the hard tickets gather dust while we crack
-through the easy ones, so we try to be mindful of this bias.
-
 ##### Priority
 
 The priority of tickets is important, and we must be biased towards dealing with
 more urgent tickets before less urgent ones.
 
-Zendesk can help us to balance these factors. When you log in and land on the
```
lexijax: Just to check that this section's gone into the ops docs? It is useful info for people on support, and there's a section on workflow that this belongs to.

playbook.md
```diff
@@ -1226,23 +1084,19 @@ But there are some limitations. Under the support service, we do not:
...
-Finally, there is an overarching principle that we don't do anything on a ticket
-that is very time consuming or that is objectively a bad idea. If you think
-you're in this situation, ask a delivery lead for advice.
-
 #### Charging for ticket work
 
 If you decide that a ticket asks for work to be done which falls outside this
 scope, then the work is chargeable. In this situation, we reassign the ticket to
-a delivery lead with an explanation.
+a delivery lead or the business development manager with an explanation.
```
lexijax: Might be better to say 'someone from the client services team' rather than biz dev manager to make sure that this isn't just one person.

playbook.md
kathcooper3: Agreed. I've only added the tailored support plan and think there's another space here for Dalmation, separate from tailored support and govpress. 

I was unclear if Dalmatian was the final name of the product, so I have left this out for now - it feels like there should be more discussion before this gets added to the playbook.

playbook.md
kathcooper3: I think we either add these to the Ops docs or keep them in the playbook. I'm hesitant to trim them down, without the support of the tech team, as these are principles that should be agreed by all. Perhaps we look to schedule a session within the next month to review these principles and update these in the playbook after that?

I've added the Client experience (AKA: ticket principles) back in, so these are unchanged for now.

playbook.md
kathcooper3: Yes, the changes we're agreeing to remove from the playbook will be moved into the ops docs, once we know what we've decided to remove.

kathcooper3:
I'm closing this PR to deal with conflicts and it was easier to create a new PR, which can be found here: https://github.com/dxw/playbook/pull/263

PR #245, opened 2019-07-30
Add a section on versioning releases to the development guide

Floppy:
So that we can predictably update code that uses them, software designed for reuse (wordpress, gems, etc) should be properly versioned.

### DIFF
--- a/guides/development-workflow.md
+++ b/guides/development-workflow.md
@@ -480,7 +480,7 @@ A new registration form and fix a bug with contacts
 
 We don't use explicit versioning for frontend projects; rather we operate a continuous deployment approach as described in the [deploying](#deploying) section.
 
-However, when building libraries, APIs, or other software components designed for reuse, versioning is important. All reusable components should have properly versioned releases. This includes Wordpress plugins and themes, ruby gems, and so on, whether open source or otherwise.
+However, when building libraries, APIs, or other software components designed for reuse, versioning is important. All reusable components should have properly versioned releases. This includes Wordpress plugins and themes, Ruby gems, and so on, whether open source or otherwise.
 
 Versioning should follow the [Semantic Versioning](https://semver.org) standard, and projects that use them should specify their requirements so that the appropriate versions are used.
 
--- a/guides/development-workflow.md
+++ b/guides/development-workflow.md
@@ -478,7 +478,7 @@ A new registration form and fix a bug with contacts
 
 ## Versioning releases
 
-We don't use explicit versioning for frontend projects; rather we operate a continuous deployment approach as described in the [deploying](#deploying) section.
+We don't use explicit versioning for application code; rather we operate a continuous deployment approach as described in the [deploying](#deploying) section.
 
 However, when building libraries, APIs, or other software components designed for reuse, versioning is important. All reusable components should have properly versioned releases. This includes Wordpress plugins and themes, Ruby gems, and so on, whether open source or otherwise.
 
--- a/guides/development-workflow.md
+++ b/guides/development-workflow.md
@@ -10,6 +10,7 @@
 - [Code review](#code-review)
 - [Code style](#code-style)
 - [Pull requests](#pull-requests)
+- [Versioning releases](#versioning-releases)
 
 ## Introduction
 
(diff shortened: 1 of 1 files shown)


guides/development-workflow.md
```diff
@@ -474,3 +475,19 @@ A new registration form and fix a bug with contacts
 - Passwords are encrypted using BCrypt, so I've taken this opportunity to
   increase the default rounds from 8 to 12
 ```
+
+## Versioning releases
+
+We don't use explicit versioning for frontend projects; rather we operate a continuous deployment approach as described in the [deploying](#deploying) section.
+
+However, when building libraries, APIs, or other software components designed for reuse, versioning is important. All reusable components should have properly versioned releases. This includes Wordpress plugins and themes, ruby gems, and so on, whether open source or otherwise.
```
lozette: Ruby should have a capital R

guides/development-workflow.md
```diff
@@ -474,3 +475,19 @@ A new registration form and fix a bug with contacts
 - Passwords are encrypted using BCrypt, so I've taken this opportunity to
   increase the default rounds from 8 to 12
 ```
+
+## Versioning releases
+
+We don't use explicit versioning for frontend projects; rather we operate a continuous deployment approach as described in the [deploying](#deploying) section.
```
lozette: Do we have a style guide for frontend vs front-end vs front end? (Should we?)

Floppy: I'm not totally happy with "frontend projects" here anyway. I mean "code that isn't reused by other code", but I couldn't think how to put it...

lozette: I'm not super-fussed tbh, but others might have an opinion 😆

erbridge: "apps", "services", "non-libraries"? I would read "frontend" are referring specifically to JS, CSS, HTML if I was reading this without full dxw context.

guides/development-workflow.md
```diff
@@ -474,3 +475,19 @@ A new registration form and fix a bug with contacts
...
+
+Versioning should follow the [Semantic Versioning](https://semver.org) standard, and projects that use them should specify their requirements so that the appropriate versions are used.
+
+Releases should be created by:
+
+- updating the version appropriately in package metadata and committing to git
+- tagging that commit using a tag of the form `vX.Y.Z`
+- building the appropriate release package from that tag
+
+Software that has not previously been versioned, but has already been used in production, should start at version `v1.0.0`.
```
erbridge: We use `v1.0.0` by default for things used in Beta projects, too? `v1.0.0` usually denotes some sort of stability.

Floppy: If there's any expectation of consistency between versions, it should be `v1.0.0` or above, but if the code hasn't yet been used in a production environment then it can start lower. The idea here was to give a baseline for all the existing things we have out there. That could probably be clearer...

guides/development-workflow.md
Floppy: I agree @erbridge, that's what I don't like about it. What about "client projects"?

pezholio: What about "consumer facing apps"? I'm trying to think of a way to express that we're talking about stuff that is on the web and people use.

erbridge: The question is "Does anything need to depend on this thing? If not, then we don't version it. Maybe it could be rephrased to just make that distinction?

guides/development-workflow.md
erbridge: I wonder if `v0.1.0` is a better default for things that are not versioned yet. `v1.0.0` seems like a big promise for something we're assigning by default.

Floppy: I agree, I can make this clearer.

* For software that is *already in use by more than one thing*, we should start at `v1.0.0`. We don't want to break existing live code accidentally, so a commitment to stability here is suitable.
* For new software, or software only being used by one thing, versioning should start at `v0.1.0`.

erbridge: I guess I'm also concerned that we're commiting to the stability of a thing without actually evaluating if it is or should be stable. I wonder if there would be some value in a note explaining what `v1.0.0` means and implies?

Floppy: Yeah, that's fair. I think actually a bit more discussion on what we mean by which versions should be a thing for the tech team forum, and I'll drop this out of here for now, pending that discussion.

guides/development-workflow.md
Floppy: The consenses of my question on twitter was that "application code" is probably the right thing, so I've changed it to that.

PR #263, opened 2019-11-01
Update to hosting and support section v2

kathcooper3:
An adjustment from previous PR to incorporate other PRs

### DIFF
--- a/playbook.md
+++ b/playbook.md
@@ -982,8 +982,7 @@ We also offer support for services that aren't built in WordPress. We are able t
 
 ### Support helpdesk
 
-The dxw Technical Operations team build, manage and run GovPress, as well as
-providing first-line support for our clients on tailored support plans.
+The dxw Technical Operations team build, manage and run GovPress, as well as provide first-line support.
 
 We use [Zendesk](https://dxw.zendesk.com) to manage support requests. Any
 incidents, requests for us to fix a problem or make a change to a site must
--- a/playbook.md
+++ b/playbook.md
@@ -971,30 +971,32 @@ to build an appropriate hosting platform for their service. This is fair
 enough, because they shouldn't have to. We built GovPress to meet this need,
 and it is now used by most of our clients.
 
+We offer [managed hosting and support](https://www.digitalmarketplace.service.gov.uk/g-cloud/services/930612236449495) for WordPress websites, as well as [blogging](https://www.digitalmarketplace.service.gov.uk/g-cloud/services/355674790119695) and [campaigns](https://www.digitalmarketplace.service.gov.uk/g-cloud/services/691308552308120) platforms, through the G-Cloud framework, configured to our client's branding requirements, all on our GovPress hosting platform.
+
 You can read more about GovPress at
 [www.govpress.com](https://www.govpress.com).
 
-We're in the process of building a new container-based hosting platform which
-will be able to support services that aren't built in WordPress. Watch this
-space.
+### Tailored support
+
+We also offer support for services that aren't built in WordPress. We are able to host and support services with our [tailored support plan for cloud applications](https://www.digitalmarketplace.service.gov.uk/g-cloud/services/393702576050692), even if your service isn't hosted with us.
 
 ### Support helpdesk
 
 The dxw Technical Operations team build, manage and run GovPress, as well as
-provide first-line support.
+providing first-line support for our clients on tailored support plans.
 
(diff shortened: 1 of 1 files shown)


playbook.md
```diff
@@ -1223,26 +1118,22 @@ But there are some limitations. Under the support service, we do not:
   Directory and suggest one, which we can then check and install.
 * Fix complex problems that come up after a plugin or WordPress core update is
   applied. Exactly what constitutes a complex problem is at our discretion. But
-  if we've spent a couple of hours on an update-related bug and it's still not fixed,
+  if we've spent a couple of hours on an updated-related bug and it's still not fixed,
```
CristinaRO: "update-related" is the correct choice here, we're using "update" as a noun (expanded, this would be "a WordPress update-related bug").

playbook.md
```diff
@@ -1177,32 +1094,10 @@ approaches and ideas if we have them - including where there's a better
 solution, but [at a cost](#charging-for-ticket-work). We always give options
 when we can.
 
-If you are in any doubt as to whether the client has understood and approved the
-change that you are about to make, explain it to them and ask for their explicit
-approval. It's fine to do this on the phone if that is more convenient, but if
-you do, update the ticket with a comment what you said and what was agreed.
-
-#### Following up on bad client feedback
```
CristinaRO: For the record, could I ask why you're removing this section? Is the information outdated, is it something we're no longer keen to have on a public page, something else?

This information could go in the commit message (I can show you how to amend it if you need/want), as a record for future readers and contributors to the Playbook.

playbook.md
```diff
@@ -1062,98 +1045,35 @@ In general:
...
-Zendesk knows each ticket's age, the time it's been since it was updated, its
-priority and the time left until we breach our SLA. Generally speaking, it makes
-good decisions about priority.
-
 Each ticket's priority should be reviewed regularly: whenever we update a
 ticket, we check the priority to make sure it's still right.
 
 #### Initial investigation
 
-When we pick up new tickets or are assigned one where no investigation has yet
```
CristinaRO: Same question: why are we removing this section? Are we moving it to the first line support guidance? I think it's useful to give as much guidance as possible for people who may be new to being assigned support tickets and dealing with them.

playbook.md
```diff
@@ -971,87 +971,70 @@ to build an appropriate hosting platform for their service. This is fair
...
 
-It is doubly bad to fail to meet a commitment and not say anything about it.
-Responsiveness is always the priority. So if for some reason we couldn't do what
-we said we'd do, we always respond to say so.
+It is doubly bad to fail to meet a commitment and not say anything about it. Responsiveness is always the priority. So if for some reason we couldn't do what we said we'd do, we always respond to say so.
 
 ##### Make a good impression
 
-In tickets as in all things, we are mindful of dxw's [values](#values).
+In tickets as in all things, we are mindful of dxw's [values](http://playbook.dxw.com/#/?id=values).
```
CristinaRO: Any reason why this is the only anchor you've expanded to an absolute link? Was it broken in a way that the other internal anchors weren't?

kathcooper3:
@CristinaRO We reviewed the support documentation in summer and had decided to remove certain sections of the support copy on the playbook, along with adding new copy to better describe our offering. 

Things that have been removed are either out of date or clients don't really need to see it.. Anything that is removed will be added to the ops documentation for internal use.

playbook.md
```diff
@@ -971,87 +971,70 @@ to build an appropriate hosting platform for their service. This is fair
...
-space.
+### Tailored support
+
+We also offer support for services that aren't built in WordPress. We are able to host and support services with our [tailored support plan for cloud applications](https://www.digitalmarketplace.service.gov.uk/g-cloud/services/393702576050692), even if your service isn't hosted with us.
 
 ### Support helpdesk
 
 The dxw Technical Operations team build, manage and run GovPress, as well as
-provide first-line support.
+providing first-line support for our clients on tailored support plans.
```
CristinaRO: The only comment I still have: to me it read more fluently with "provide", as in "The [...] team build, manage and run [...], as well as provide". "The team" is the subject, and "build, manage, run, provide" are all actions done by the subject. But I'm not a native speaker, so maybe this is one of those cases where what "sounds" right isn't right grammatically. 🤷‍♀

kathcooper3: Hmm, that still doesn't sound quite right. Let me revert this change to the original sentence, without mentioning clients on tailored support plans.

PR #265, opened 2019-11-08
Switch to Github actions workflow for Vale

mec:
We use [Vale](https://github.com/errata-ai/vale/) to lint the playbook.

Before this PR we ran Vale via Travis, which had started to fail. We are also using Github actions whenever possible so it was a good time to switch the Playbook over.

We run Vale on `contributing.md`, `README.md`, `playbook.md` and everything in `guides` except `style-guide.md` as it contains examples of the styles we do not want to see.

We used the [official Vale action](https://github.com/errata-ai/vale-action) for this which can only take a single argument to specify which files to lint.  Fortunately Vale allows you to disable the linting in the file itself using `<!-- vale off -->`. This allowed us to skip linting `style-guide.md`, we could go one step further and wrap the examples of poor style in `<!-- vale off -->` and `<!-- vale on -->` if we wanted to.

The Vale action has some great features like warnings in the Github "Files changed" UI.

We removed `.travis.yml` to stop the old workflow running and once this PR is merged, we can remove any webhooks for Travis.

As Vale warnings now show up in Github, there are some content things we may want to fix:

Screenshot_2019-11-08 Switch to Github actions workflow for Vale by mec · Pull Request #265 · dxw playbook

🙀


erbridge:
I think this could do with some documentation about how to ignore things Vale gets wrong. For example, the `deliver` in the first annotation is used in the way they use it in the example.

guides/style-guide.md
```diff
@@ -1,3 +1,5 @@
+<!-- vale off -->
```
erbridge: I would argue that this document could be used as an example of how to turn off specific warnings well, in a way that doesn't turn them off for other bits of text. Does Vale support only disabling a specific rule?

.github/workflows/linting.yml
```diff
@@ -0,0 +1,16 @@
...
+    steps:
+    - name: Checkout
+      uses: actions/checkout@master
+
+    - name: Vale
+      uses: errata-ai/vale-action@v1.0.1
+      env:
+        GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+      with:
+        config: https://raw.githubusercontent.com/dxw/playbook/master/.vale.ini
```
erbridge:
```suggestion
        config: .vale.ini
```

Can't you do that, allowing changes to the Vale config on branches to run that changed config?

mec: Looks like `.vale.ini` is included by default so we can loose the `config` althogether.

erbridge: :+1:

guides/style-guide.md
mec: I think it's interesting and begs the question, what are these rules for? 

I don't think Vale is smart enough to really help out, like the `deliver` example you mention, it's really more a que to the user writing the content to check their use of `deliver` is the right one, see: 

https://github.com/dxw/playbook/blob/c905e043f11fcc18e11adbd85f26c3f24b6f7be1/styles/govuk/words-to-avoid.yml#L12

It looks like Vale has a `suggestion` setting for this, but Github won't surface the suggestions so it would not help a lot of the people who create the content for the Playbook. I assume none of these warnings have ever been seen by folk not running Vale locally?

In terms of disabling, it looks like you can change the level of the error Vale raises but not skip a specific rule, I might be wrong though:

https://errata-ai.github.io/vale/config/#ignoring-parts-of-your-documents

erbridge: This is what I was expecting:

```md
<!-- vale Style.Rule = NO -->

This is some text

<!-- vale Style.Rule = YES -->
```

which would disable `Style.Rule` for the text in the block.

> I assume none of these warnings have ever been seen by folk not running Vale locally?

I would think that's right, but they will _start_ to see them with this change, which means we need to be able to tell people how to fix them. I think having rules are useful (see Sensitivity Bot), but allowing people to override them, and say that they have read the warning, but think they are right, regardless, is a good thing. It also documents that someone has _decided_ that their choice of words is intentional.

styles/govuk/words-to-avoid.yml
```diff
@@ -1,7 +1,7 @@
 extends: substitution
 message: '%s'
 link: https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#words-to-avoid
-level: warning
+level: suggestion
```
erbridge: :+1:

guides/style-guide.md
mec: Agreed, I guess I am not comfortable to add an extra layer of complexity, no matter how small, to getting changes into the Playbook without consultation.

As these are only warnings, or suggestions now, they don't block merging the content into the playbook so a user can still choose to ignore them through that process?

I also cannot seem to get `<!-- vale Style.Rule = NO -->` to work, tired `<!-- vale Style.Rule = NO -->` and specific ones like `<!-- vale dxw.tired-idioms = NO -->` which is what I had assumed the syntax was.

mec: All I can think of is we don't follow the CamelCase rules explicitly for our styles, so it should be:

`<!-- vale Dxw.TiredIdioms = NO -->`

mec: Yep, looks like the camelcase is required and it's a bit sensitive to where the comments are. I'll have to look at renaming the styles and rules.

mec:
I have to jump off this now, but I'll try and pick it up on Monday.

mec:
I've updated all the style and rules filenames and it all seems to be working. I am not loving the spacing that seems to be required for the switches to work.

guides/style-guide.md
```diff
@@ -45,47 +45,77 @@ Do this:
...
+<!-- vale Dxw.GCloud = YES -->
+
+<!-- vale Dxw.GovPress = NO -->
+
 **GovPress**: not 'Govpress' or 'Gov Press'
 
+<!-- vale Dxw.GovPress = YES -->
+
+<!-- vale Dxw.Multisite = NO -->
+
```
erbridge: I sort of disagree. I'm fine with this spacing, tbh.

Floppy:
LGTM! Merge when ready, @mec :)

PR #268, opened 2019-11-20
Guidance on taking notes and recording

johnwaterworth:
General good practice and specific guidance on using different tools and devices for different kinds of recordings.
Guidance on recording phone calls to come once we've agreed it.
Purchase request in for compact cameras.


johnwaterworth:
Thanks Vita. I've added a section on using sticky notes and paper worksheets, along with some other improvements.

All the points starting with 1. is how numbered lists work in Markdown.

PR #291, opened 2019-12-20
Update to 'reporting our time' section

kathcooper3:
Following the new process in which all team members are now required to add their approved planned leave into 10,000ft, this has been reflected in the playbook.

### DIFF
--- a/playbook.md
+++ b/playbook.md
@@ -2016,7 +2016,7 @@ Everyone who works on a client project at dxw digital is responsible for reporti
 
 We use [10,000ft](https://app.10000ft.com/) to schedule our work, which helps us to forecast our capacity over future weeks and months in a reliable way.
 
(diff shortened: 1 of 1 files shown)


playbook.md
```diff
@@ -2016,7 +2016,7 @@ Everyone who works on a client project at dxw digital is responsible for reporti
 
 We use [10,000ft](https://app.10000ft.com/) to schedule our work, which helps us to forecast our capacity over future weeks and months in a reliable way.
 
-Our Commercial Operations team update the schedule to reflect any planned leave, which helps us to look ahead - this is recorded as “Out of office". Each member of our delivery team uses the tool to record the days they work on client projects on a weekly basis.
+When team members planned leave has been approved, team members are responsible for updating their schedule in 10,000ft to reflect this. This helps us to look ahead and is recorded as “Out of office". Each member of our delivery team uses the tool to record the days they work on client projects on a weekly basis.
```
CristinaRO: I think this would work better, and be more consistent, if it were using the second person.

The other instructions are addressed to the reader as "Add *your* holiday", "any event *you* attend", so I'd prefer to keep the style consistent as "When your leave has been approved [...]".

kathcooper3: @CristinaRO I've changed this to reflect your feedback.

PR #293, opened 2019-12-30
Updates on taking leave

johnwaterworth:
Updates to the Holiday section and new Compassionate leave section


johnwaterworth:
Agree about including Becs. At the moment she doesn't have access to Github. I suggested she gets an account when she gets back from holiday.

PR #301, opened 2020-02-04
explainer for design career progression framework

gazaston:
Adding a guide to the playbook for the design career progression framework. Just the basics for now and very much subject to change as we learn more about what works and what needs improving.


gazaston:
[Document the operation of the designers progression framework](https://trello.com/c/FpgdPtSt/80-document-the-operation-of-the-designers-progression-framework)

PR #305, opened 2020-02-10
add a better explaination of how review meetings work

gazaston:
- including the new policy on frequency of reviews


gazaston:
[Codify (Playbook?) the meetings a direct report can expect to have with their line manager](https://trello.com/c/hzKqq2eM/61-codify-playbook-the-meetings-a-direct-report-can-expect-to-have-with-their-line-manager)

PR #307, opened 2020-02-12
add text about third parties in formal reviews

gazaston:
- be explicit about the option to involve a third party where required


guides/design-career-progression-framework.md
```diff
@@ -62,6 +62,8 @@ In these meetings you’ll review progress on your chosen proficiencies over the
 
 Every member of the team has their own Google Sheet, with a tab dated for each formal review, where levels of proficiency are tracked and displayed as charts. These sheets are the single source of truth for an individual's progression, and they're filled out with your manager in the review meeting.
 
+If you and your manager can't reach agreement about something in a progression review, or your progression in general, either of you have the option of involving a third party.
```
titlescreen: Sounds like a good improvement. Does this need some context as to what constitutes a third party? Would it be a workmate, person from HR or someone external to the business?

gazaston: @titlescreen All of those are valid I reckon.

PR #308, opened 2020-02-14
Add files for 10kft guide

kathcooper3:


Floppy:
Might rename the files later, but that doesn't need to be a blocker.

PR #313, opened 2020-02-14
Improve vale linting on CI

Floppy:
Vale is lovely, but leaves a lot of noise on PRs, because it includes sometimes-inappropriate suggestions.

So that we can trust it again and not ignore it, I've changed it to only show warnings and errors by default. There is a new npm script to run the full thing with suggestions as well. This should make our PRs quieter, and mean we can enforce the style guide properly as a required test on PRs.


CristinaRO:
Will this also get rid of GitHub forcing "Unchanged files with check annotations" to show up, thus making the diff even harder to evaluate? 🙏

PR #314, opened 2020-02-14
Extreme prose linting

Floppy:
Added `proselint` and `write-good` styles to vale to help write English good. Builds on top of #313

### DIFF
--- a/.vale.ini
+++ b/.vale.ini
@@ -4,7 +4,7 @@ StylesPath = _styles
 
 MinAlertLevel = warning # suggestion, warning or error
 
-BasedOnStyles = Dxw,Govuk
+BasedOnStyles = Dxw,Govuk,proselint,write-good
 
 # Style.Rule = {YES, NO, suggestion, warning, error} to
 # enable/disable a rule or change its level.
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,6 @@
+[submodule "_styles/proselint"]
+	path = _styles/proselint
+	url = git://github.com/errata-ai/proselint
+[submodule "_styles/write-good"]
+	path = _styles/write-good
+	url = git://github.com/errata-ai/write-good.git
--- a/.prettierignore
+++ b/.prettierignore
@@ -8,6 +8,9 @@
 !/.github
 !/.prettierrc
 /_site/**/*
+/_styles/**/*
+!/_styles/dxw/
+!/_styles/Govuk/
 /.jekyll-cache/**/*
 /dist/**/*
 /script/**/*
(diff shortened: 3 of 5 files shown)


.vale.ini
```diff
@@ -1,9 +1,9 @@
 StylesPath = styles
-MinAlertLevel = suggestion # suggestion, warning or error
+MinAlertLevel = warning # suggestion, warning or error
 # Only Markdown and .txt files; change to whatever you're using.
 [*.{md}]
 # List of styles to load.
-BasedOnStyles = Dxw,Govuk
+BasedOnStyles = Dxw,Govuk,,write-good,proselint
```
CristinaRO: Is there an invisible style between `Govuk` and `write-good`?

Floppy: you'd think the linter would catch that :D

erbridge:
@Floppy Why is this in draft? Is there stuff left to do?

lawrence-forooghian:
I don't understand what this does. I'm in favour of anything that helps us write better, but as long as it doesn't get in the way. I'd like to understand how opinionated it is and how we can override it. Could you give some examples of the sorts of errors / "errors" that it can catch?

rjw1:
I'm not entirely keen on submodules but for this use its probably okay.

Floppy:
I stalled on this and never got back to it. It was to add some more style checks to help make sure we have a consistent tone. It would be pass/fail for serious issues, and would be there by default, though much of what it checks for is in the warning or suggestion category. It’s probably too much, but I’m not sure.

As for what it checks for, see https://github.com/amperser/proselint/#checks and https://github.com/btford/write-good#checks. There are things like weasel words, passive voice, removing unnecessary words, etc, all of which should help make tighter clearer written content.

Most of what they check for come up as warnings or suggestions, which won't block the build, but there are a few things it would flag as errors, such as starting a sentence with "So", or use of gender biased words (https://github.com/errata-ai/proselint/blob/38232268357ca0b43c35d36e24364b730da27185/proselint/GenderBias.yml).

I think it's worth a revisit if anyone has time and the inclination to learn about `vale`, but if it's just sitting around not doing anything then equally it could be closed off until resurrected at some point in the future.

erbridge:
How about we turn it on and remove it if it's annoying. Because it's Vale, comments get added as annotations to PRs, so there will be direct feedback to submitters.

CristinaRO:
If it was only annotating new / changed content, it would be ok. But due to this "feature": https://developer.github.com/changes/2019-09-06-more-check-annotations-shown-in-files-changed-tab/ it will pester endlessly about all files in the repo, and make "Files changed" harder to use for reviewers.

erbridge:
Fair, but it currently doesn't find any issues, so I think that's ok?

PR #330, opened 2020-04-03
Make a Who we are section

johnwaterworth:
Move existing material into a Who we are section (see the proposed structure in the Playbook content strategy)

With existing content in this new structure we can make further updates:
 - update the Why we're here mission statement
 - add a people-centred principle

### DIFF
--- a/playbook.md
+++ b/playbook.md
@@ -1,36 +1,8 @@
 # dxw’s Playbook
 
-## Purpose
+## Who we are
 
-This Playbook is our reference for who we are and the way we do things. Something canonical that tells us what the current “right way” to do things is.
-
-If you’re a current or potential client, this Playbook is also for you. To help you understand us and how we can work together.
-
-We continually update the principles and guides in our Playbook. Because we’re always exploring better ways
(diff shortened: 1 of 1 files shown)


playbook.md
```diff
@@ -168,6 +140,34 @@ regularly assessing the impact of all our policies and practices.
 
 Thank you for being you - it makes us better.
 
+### Advisory Council
+
+dxw does not have external investors or shareholders. To help keep us honest, we are very lucky to have a voluntary advisory council, made up of a small group of senior figures, who meet quarterly to discuss and provide guidance on our strategy, business plan and other topics related to company health and progression. The group have a wealth of experience of public and private sector work, as well as being friends of dxw that share our values.
+
+### This Playbook
```
erbridge, line 147:
```suggestion
## This Playbook
```

I think this belongs at a level up?

johnwaterworth: I remember debating this last year. I'd prefer to leave the bit about the Playbook under the Who we are section rather than have a top-level section all of its own.

erbridge: It's not where I'd look for information about the playbook, but I guess we can hash that out separately.

PR #339, opened 2020-05-01
Fix the casing on a heading

erbridge:


lawrence-forooghian:
@erbridge please can you fix the conflict?

PR #351, opened 2020-05-27
Use the same favicons as the dxw website

CristinaRO:


CristinaRO:
[Add favicons to Playbook](https://trello.com/c/hAERFUsX/59-add-favicons-to-playbook)

PR #355, opened 2020-06-03
Use Jekyll for Playbook

pezholio:
Fixes #349 

Until now, we've used [Docsify](https://github.com/docsifyjs/docsify) to power the Playbook, which has worked OK, but it doesn't work at all with JavaScript turned off. We wouldn't accept this in client projects, so it makes sense for the Playbook to meet the same standards.

I've tweaked the Playbook to use [Jekyll](https://jekyllrb.com/) to build the site, as well as using Webpack to build the limited Javascript functionality. Unfortunately, as the Playbook is a static site, the search won't work without Javascript, but I think this is a fair trade-off. 

I've got this branch deploying to [Netlify](https://app.netlify.com), as it gives us more flexibility in how we deploy the site. Assuming we're OK with this, it'll need a bit more configuration for us to move it across.

You can see the branch deployed the Netlify here - https://competent-bhabha-662efd.netlify.app/

PR #365, opened 2020-06-12
Switch to Jekyll over Docsify

erbridge:
Docsify relies on JavaScript. This changes to using Jekyll to build the site. We then automatically deploy the built site to GitHub pages.

This will automatically commit any autofixable issues to the branch after testing. It uses my personal access token to do that, with the minimal permissions. Probably it would be better to use a service account or something instead.

Replaces #355.

## To do after merge:

- [x] Switch the GitHub Pages branch setting to `gh-pages`
- [ ] Turn off Hound entirely and remove the config from the repo
- [ ] Change the branch protection rules to protect `gh-pages` from deletion and require the new CI workflows to pass

## Other things:

We should really by autoprefixing our CSS. We probably want to add PostCSS into our build pipeline.


_includes/navigation.html
```diff
@@ -0,0 +1,20 @@
...
+      {% for p in pages %}{% if p.path contains 'guides/' and p.path !=
+      page.path %}
+      <li><a href="{{ p.url }}">{{ p.title }}</a></li>
+      {% endif %}{% endfor %}
+    </ul>
+  </li>
+</ul>
+
+<div class="logo">
+  <a href="/" title="Back to playbook home page">
```
CristinaRO: Should this be "Playbook" to be consistent with the capitalisation changes elsewhere?

_layouts/default.html
```diff
@@ -0,0 +1,73 @@
...
+                stroke-width="2"
+                d="M8.25,15.75,15.75,8.25"
+              ></path>
+            </svg>
+          </div>
+        </div>
+        <div class="results-panel"></div>
+      </div>
+      <h1 class="app-name">
+        <a class="app-name-link" data-nosearch="" href="/">dxw’s playbook</a>
```
CristinaRO, line 43: Same: should this be "Playbook"?

guides/how-we-do-user-research.md
```diff
@@ -239,9 +238,9 @@ recordings, or deleting data from prototypes.
 We have template discussion guides for three different kinds of research
 activities:
 
-- [Interviews and contextual research](https://drive.google.com/open?id=1XeBhmQL0jqFFsl2Cw-DOKVpabTRILGl-TjeyUDW90vM)
-- [Usability tests](https://drive.google.com/open?id=1hcVkLGJVDB9ZOoMIz93erUpk1RUgXN_ZcZflJxsFpHw)
-- [Workshops](https://drive.google.com/open?id=1ZHZwLBnUlDsxrvv1qj_ZpFU9QGsP_dIDLMVcGbEYswY)
+- [Interviews and contextual research](https://drive.google.com/open#1XeBhmQL0jqFFsl2Cw-DOKVpabTRILGl-TjeyUDW90vM)
```
CristinaRO, line 241: Ah, this looks like a regexp catching Google Drive links in its net, here and everywhere in this file.

guides/index.html
```diff
@@ -0,0 +1,11 @@
+---
+title: Guides
+---
+
+{% assign pages = site.pages | sort: "title" %}
+
+<ul>
+  {% for p in pages %}{% if p.path contains 'guides/' and p.path != page.path %}
+  <li><a href="{{ p.url}}">{{ p.title }}</a></li>
```
CristinaRO, line 9: Should the lack of a space before the first set of closing braces have been caught by the various linters or formatters?

guides/managing.md
```diff
@@ -2,125 +2,164 @@
...
 For more significant additions and changes to the Playbook, the steps are:
 
- - the owner of the section and the author agree the need for the update
- - the author drafts and reviews the update with appropriate colleagues
- - the author follows the [guide to contributing to the Playbook](https://playbook.dxw.com/#/contributing)
- - the owner and colleagues review and approve the changes
+- the owner of the section and the author agree the need for the update
+- the author drafts and reviews the update with appropriate colleagues
+- the author follows the
+  [guide to contributing to the Playbook](https://playbook.dxw.com/#/contributing)
```
CristinaRO, line 101: Should this become a `/guides/` link instead of using the octothorp? ( ;) )

guides/team-and-project-scheduling.md
```diff
@@ -91,11 +91,9 @@ The **project lead** should:
 ### Reporting our time
 
 Every team member working on a client or named project is responsible for
-reporting their time, as mentioned
-[in the Playbook](http://playbook.dxw.com/#/?id=reporting-our-time). More on how
-to do this in the
-[reporting time](http://playbook.dxw.com/#/guides/team-and-project-scheduling?id=reporting-time)
-section of this guide.
+reporting their time, as mentioned [in the Playbook](/#reporting-your-time).
+More on how to do this in the [reporting time](#reporting-time) section of this
```
CristinaRO, line 95: Note to self to check if reporting your time is still a separate guide as well as an internal anchor link.

_layouts/default.html
erbridge: Thought I'd got them all. Should be fixed now. Thanks.

guides/how-we-do-user-research.md
erbridge: Whoops. I'll fix. Thanks.

guides/index.html
erbridge: The linter doesn't actually understand Liquid (Jekyll's templating language), but it's HTML compliant so it does the right thing most of the time.

guides/managing.md
erbridge: Yup. I caught that too :).

.browserlistrc
```diff
@@ -0,0 +1,3 @@
+> 5%
+last 2 versions
+IE 11
```
erbridge, line 3: We should think about whether these are the correct browsers.

guides/tone-of-voice.md
```diff
@@ -1,103 +1,155 @@
 # Tone of voice
```
CristinaRO, line 1: Does this *need* the formatting to start with `---` like the other files?

guides/managing.md
```diff
@@ -2,125 +2,164 @@
 
 This is the content strategy for dxw Playbook.
```
CristinaRO, line 4: This file probably needs the `---` structure at the top as well. None of the linters or tests are catching it, so this is just a guess.

CristinaRO:
[Make the playbook work without JavaScript enabled](https://trello.com/c/FfN5pJwe/49-make-the-playbook-work-without-javascript-enabled)

CristinaRO:
[Resolve accessibility issues on Playbook](https://trello.com/c/bWWTHrwS/50-resolve-accessibility-issues-on-playbook)

erbridge:
The `---\n---` is called frontmatter, FYI, and it's needed for Jekyll to process it correctly.

erbridge:
> Either the linters or the tests to catch the missing heading in files

This is probably a good idea. Not sure any do it by default as Markdown files without frontmatter are still valid in Jekyll, just not categorized as pages (so not found by the code that loops through all the pages). We might have to write a script ourselves that searches for it based on path or something.

guides/support-and-on-call.md
```diff
@@ -1,4 +1,6 @@
-# Support and on call
+---
+title: Support and on call
+---
 
 We have service definitions for [GovPress hosting], [container hosting] and
```
CristinaRO, line 5: These don't turn into links in the HTML-ised version.
Screenshot 2020-06-15 at 16 32 39

erbridge: Oh, interesting...

PR #382, opened 2020-06-18
Improve general styling and change the colour scheme to be a little easier on the eyes

erbridge:
image


leeky:
@erbridge Personally I find the bright teal sidebar a little harder to read than the current dark green version, but it could be because I'm viewing it on calibrated screens 🤷

CristinaRO:
@erbridge Yeah, I'm sorry, it's the teal in large amounts on the screen that is overwhelming, and coupled with orange or yellow it only makes it harder.

I would try blue and yellow myself if I had a spare moment to play with the dev tools: https://v-df7640e636dddb2c9710.brand.madebyfieldwork.com/section/core-elements#colour-palette

erbridge:
I'll switch to another. This is pretty easy now.

erbridge:
How's this?

image

CristinaRO:
Screenshot 2020-06-19 at 12 53 24
or (my favourite so far)
Screenshot 2020-06-19 at 12 57 11
or even
Screenshot 2020-06-19 at 13 02 15

erbridge:
I've been trying to use one that has a square avatar set up.

erbridge:
I like the cream one actually. Let me try that.

erbridge:
Ok, let's try again:

image

gazaston:
> Ok, let's try again:
> 
> image

Have we contrast-checked this?

erbridge:
Yup. Passed Firefox's contrast check.

erbridge:
300 is a very light font-weight though for Poppins.

erbridge:
I've made to 400 again, even though it disagrees with the brand guide.

image

gazaston:
Screenshot 2020-06-19 at 14 19 47

Did FF think this was okay, or perhaps it doesn't check placeholders?

erbridge:
It did. I've upped the contrast on it.

image

_layouts/default.html
```diff
@@ -21,21 +21,6 @@
             aria-label="Search text"
             placeholder="Type to search"
           />
-          <div class="clear-button">
```
CristinaRO, line 24: I would like to understand better why we're removing the clear search button without a replacement. I don't fully understand what overflowing the layout means.

Is it a functionality that can "just" be done without?

erbridge: We can add it back in, but it causes the sidebar to overflow currently, and I'm tired of fighting Docsify's defaults.

erbridge: And I don't think it provides much value.

PR #383, opened 2020-06-18
Exclude the Guides index page from the menu

erbridge:


pezholio:
Good catch. Missed that 👍🏻

PR #390, opened 2020-06-19
Move the sidebar to side under the navbar on large screens

erbridge:
We don't do it for small screens as they're not likely to have a lot of vertical real estate.

image

### DIFF
--- a/_stylesheets/styles.scss
+++ b/_stylesheets/styles.scss
@@ -58,10 +58,11 @@ strong {
 }
 
 .content {
-  padding-top: 95px;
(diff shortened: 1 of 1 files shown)


johnwaterworth:
This is looking really good. Thank you for trying lots of colour options.

My only suggestion is to keep the Playbook and Guides links text-aligned right on both large and small screens. They should be easier to see and find up there.

erbridge:
Can do.

erbridge:
Better @johnwaterworth?

image

johnwaterworth:
That's great

PR #397, opened 2020-06-26
Use conditionals to avoid checks from being caught by `set -e`

erbridge:
This fixes the installation scripts for first time users.


erbridge:
Replaced by #398.

erbridge:
That's a lie. Wrong PR...

PR #398, opened 2020-06-26
Update the introduction to Who we are

johnwaterworth:
Update the introduction to Who we are to reflect the nw things we say about ourselves


johnwaterworth:
I still can't pick Louise as a reviewer though.

PR #408, opened 2020-07-10
Add junior developer plan as a guide

erbridge:


erbridge:
@MariluzHerrera You're welcome :)

PR #418, opened 2020-07-31
Add more info about oncall escalation and TOIL.

rjw1:


guides/support-and-on-call.md
```diff
@@ -159,3 +159,36 @@ Outside office hours, the expected response depends on the severity of the
 incident. We have a set of factors that determine the severity of an incident,
 and have included these in our
 [Service Level Agreement](https://contracts.dxw.com/service_level_agreement.md).
+
+### On call escalation
+
+While you are on out of hours on call there is also a 2nd line on call rota
```
erbridge, line 165:
```suggestion
While you are on call out of hours there is also a 2nd line on call rota
```

Nitpick :)

guides/support-and-on-call.md
```diff
@@ -159,3 +159,36 @@ Outside office hours, the expected response depends on the severity of the
...
+will alert them if you do not acknowledge an alert within 5 minutes. This is to
+make sure that alerts are dealt with in a timely manner. You should not feel bad
+about missing an OpsGenie alert. This is one of the reasons why we have the
+second line rota.
+
+You can also use OpsGenie to escalate an alert to them if you are unsure what to
+do, lack access or need a hand in general. If it is a major incident you may
+want to escalate so that they can deal with client communications. If the person
+on the 2nd line rota does not respond feel free to contact the Head of Technical
+Operations or the CTO to help you. You should also feel free to contact other
```
erbridge, line 176: How do I contact those two people out of hours when I'm bleary and not thinking straight?

guides/support-and-on-call.md
```diff
@@ -159,3 +159,36 @@ Outside office hours, the expected response depends on the severity of the
...
+BreatheHR
+
+- alerts while you are awake you can claim 0.25 days of TOIL
+- alerts which wake you up you can claim 0.5 days of TOIL
+
+If you have to work more than an hour fixing an alert you can claim extra TOIL
+at 0.25 days per hour.
+
+If you have been woken up by an alert you are not expected to start work at your
+normal time. If you feel that you need a rest from our of hours on call after a
```
erbridge, line 192:
```suggestion
normal time. If you feel that you need a rest from out of hours on call after a
```

erbridge, line 192:
```suggestion
normal time. If you feel that you need a rest from being on call out of hours after a
```

guides/support-and-on-call.md
evilstreak: My phone number is on my Slack profile, but the only thing that'll make it through do-not-disturb settings in the middle of the night are calls from one of the OpsGenie numbers. Might need to look at escalation routes within OpsGenie?

guides/support-and-on-call.md
```diff
@@ -159,3 +159,36 @@ Outside office hours, the expected response depends on the severity of the
...
+do, lack access or need a hand in general. If it is a major incident you may
+want to escalate so that they can deal with client communications. If the person
+on the 2nd line rota does not respond feel free to contact the Head of Technical
+Operations or the CTO to help you. You should also feel free to contact other
+members of the team if the incident is serious enough and you have discussed it
+with the person you have escalated to.
+
+### Claiming TOIL for out of hours alerts
+
+If you are alerted out of hours you can claim Time Off in Lieu(TOIL) in
```
evilstreak, line 182:
```suggestion
If you are alerted out of hours you can claim Time Off in Lieu (TOIL) in
```

guides/support-and-on-call.md
```diff
@@ -159,3 +159,36 @@ Outside office hours, the expected response depends on the severity of the
...
+BreatheHR
+
+- alerts while you are awake you can claim 0.25 days of TOIL
```
evilstreak, line 185: There's a rules-lawyer bit of me looking at this and seeing giant holes: 12 alerts in the space of 30 minutes is not the same as one alert per hour for most of a day.

I'm struggling to word this at all clearly, but I think something around disrupted hours would be more about the time spent and frequency of alerts, and less about the raw volume of alerts generated.

Basically: imagine that when the first alert hits you start a 1 hour timer and claim some TOIL. If you're still working when that timer finishes, restart it and claim more TOIL. Once you're done, and a new alert brings you back, see if the timer is still running. If it isn't, start it again and claim more TOIL.

Does that make any sense?

evilstreak: (In case it's not clear, which on re-reading I think it may not be: I'm not saying we should get people to start actual timers, just trying to use a hypothetical timer to illustrate the approach)

rjw1: yeah, that makes sense (I was trusting people to not game the system). ill think of some words :)

guides/support-and-on-call.md
rjw1: it can almost certaihly be escalated somehow also you can use the directory in opsgenie to get our numbers.

rjw1: ive found a way which i can document.

rjw1:
@evilstreak @erbridge I've updated this PR with your suggestions.

erbridge (approved):

PR #447, opened 2020-09-18
Ignore docsify in Dependabot

erbridge:
We only use it for styling now, until we replace it with our own styles, so we don't want updates in case they break our overrides.


Gweaton (commented):
Has "fix formatting" snuck in from https://github.com/dxw/playbook/pull/446?

erbridge:
Yes, or otherwise the tests wouldn't pass. It's the same commit, so won't mess up history.

Gweaton (approved):

PR #471, opened 2020-10-02
Add "Getting a project ready for support" guide

jacksonj04:
This explains what needs to happen at the end of a project to get it ready to be supported in more detail. It replaces and expands on the bullet point as part of the wrapping-up guide.

### DIFF
--- a/guides/getting-a-project-ready-for-support.md
+++ b/guides/getting-a-project-ready-for-support.md
@@ -1,0 +1,75 @@
+---
+title: Getting a project ready for support
+---
+
+For some projects, once the primary development phase has wrapped up dxw will
+continue to offer support (subject to a support contract). Since we rotate
+support staff, things need to be in place to let them effectively respond to
+issues.
+
+## Make sure the documentation is up to date
+
+Support developers may not have context on the project, and rely on the
+documentation to effectively solve problems. For most projects:
+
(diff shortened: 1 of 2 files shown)


mattymaxwell (approved):

guides/getting-a-project-ready-for-support.md
```diff
@@ -0,0 +1,75 @@
...
+should include either documentation or (preferably) links to documentation for
+any on-boarding steps needed, and the front line support runbook.
+
+## Set up monitoring systems
+
+Make sure that enough monitoring systems are in place to raise warnings in case
+of outages, errors or service degradation (the exact level of monitoring will
+depend on the project and support contract).
+
+### Make sure error reports are raised in Zendesk
```
CristinaRO, line 72: Not a blocker for this version, but in future iterations it would be nice to have a reminder that teams should at least *try* and filter error reports to those that are actually useful and actionable by the support developer, otherwise Zendesk is going to be flooded with Rollbar errors that require context before even knowing whether there is anything to do about them.

This should be on the roadmap much earlier, since it's unlikely any team will be able to do all the error trimming in the last sprint, but we can try to raise awareness, at least.

CristinaRO (approved):
Thank you for writing this up!

PR #476, opened 2020-10-06
Front matter always present for guides

LornaHarwood:
`jekyll_frontmatter_tests` gem has been implemented to check that each markdown page contains a front matter, with a title present.
This test has been added to CI tests to ensure this check happens automatically.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,3 +4,8 @@ gem "foreman"
 gem "jekyll"
 gem "nokogiri"
 gem "standard"
+
+group :jekyll_plugins do
+  gem 'pry'
+  gem 'jekyll_frontmatter_tests'
+end
--- a/Gemfile
+++ b/Gemfile
@@ -6,6 +6,6 @@ gem "nokogiri"
 gem "standard"
 
 group :jekyll_plugins do
-  gem 'pry'
-  gem 'jekyll_frontmatter_tests'
+  gem "pry"
+  gem "jekyll_frontmatter_tests"
 end
--- a/deploy/tests/schema/_posts.yml
+++ b/deploy/tests/schema/_posts.yml
@@ -3,7 +3,7 @@ title: "String"
 config:
   path: "guides"
   ignore:
-  - "README.md"
-  - ..
-  - .
-  - .DS_Store
+    - "README.md"
+    - ..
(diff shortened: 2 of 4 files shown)


Gemfile
```diff
@@ -6,6 +6,6 @@ gem "nokogiri"
 gem "standard"
 
 group :jekyll_plugins do
-  gem 'pry'
-  gem 'jekyll_frontmatter_tests'
+  gem "pry"
```
lozette, line 9: This is the first time I've looked at a PR in this repo since actions were added, but do you think it would be possible to squash this commit into the relevant earlier commits where these files were added?

lozette: I've had a look at other PRs and it looks like automatic fixes are a thing with their own commits, so you can ignore me!

lozette (approved):

script/test
```diff
@@ -37,5 +37,8 @@ else
   npm run lint
 fi
 
+echo "Testing Front Matter"
```
erbridge, line 40:
```suggestion
echo "==> Checking for required frontmatter..."
```

For consistency.

deploy/tests/schema/_posts.yml
```diff
@@ -0,0 +1,8 @@
+title: "String"
+config:
+  path: "guides"
+  ignore:
+    - "README.md"
+    - ..
+    - .
```
erbridge, line 7: Are these really necessary? Kind of wild if the plugin follows these paths.

deploy/tests/schema/_posts.yml
```diff
@@ -0,0 +1,8 @@
+title: "String"
+config:
```
erbridge, line 2:
```suggestion
title: "String"

config:
```

Pedantic I know, but I think it helps clarify the fact that `title` is a constraint and `config` is general configuration, and they're not the same type of thing.

erbridge (approved):
Non-blocking suggestions.

deploy/tests/schema/_posts.yml
tahb: I can't remember exactly what the gem is doing but we certainly found out the hard way that these are required after trying to exclude them to begin with.

PR #491, opened 2020-10-19
Set focus correctly

pezholio:
When we're scrolling to the right part of the page when using the sidebar navigation, the focus remains on the link, which can be confusing for screenreader users. Because `window.scrollTo` doesn't have a callback function, we need to use `setInterval` to poll to see if the page has been scrolled to roughly the right place. If it has, we remove focus from the clicked link, and set focus to the target heading element.

### DIFF
--- a/_javascript/lib/sidebarNavigation.js
+++ b/_javascript/lib/sidebarNavigation.js
@@ -10,22 +10,38 @@ export default (links) => {
     return element.getBoundingClientRect().top + window.pageYOffset - yOffset;
   };
 
-  const scrollToLink = (element) => {
-    const locationID = element.getAttribute("data-target");
+  const scrollToLink = (link) => {
+    const locationID = link.getAttribute("data-target");
     const targetElement = document.getElementById(locationID);
+    const y = getElementLocation(targetElement);
 
     window.scrollTo({
-      top: getElementLocation(targetElement),
+      top: y,
       behavior: "smooth",
     });
 
-    if (window.innerWidth < 768) {
-      const body = document.querySelector("body");
+    const intervalID = setInterval(function () {
+      // Check if the Y Offset has got within 10 pixels of the target
+      // it's not always going to be exact, so make sure we're roughly
+      // there
+      if (Math.abs(y - window.pageYOffset) < 5) {
+        link.blur();
+        // Set focus on the target element. Because it's not an anchor
+        // or form element, we need to give it a tabindex first
+        targetElement.setAttribute("tabindex", "-1");
+        targetElement.focus();
 
-      body.classList.remove("close");
-    }
+        if (window.innerWidth < 768) {
+          const body = document.querySelector("body");
 
-    history.pushState({}, "", "#" + locationID);
+          body.classList.remove("close");
+        }
+
+        history.pushState({}, "", "#" + locationID);
+
+        clearInterval(intervalID);
+      }
+    }, 300);
   };
 
   for (const link of links) {
--- a/_spec/javascript/lib/sidebarNavigation.test.js
+++ b/_spec/javascript/lib/sidebarNavigation.test.js
@@ -9,6 +9,7 @@ describe("sidebarNavigation", () => {
     window.scrollTo = scrollToMock;
 
     jest.spyOn(history, "pushState");
+    jest.useFakeTimers();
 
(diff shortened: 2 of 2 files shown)


_javascript/lib/sidebarNavigation.js
```diff
@@ -13,19 +13,39 @@ export default (links) => {
...
 
     window.scrollTo({
-      top: getElementLocation(targetElement),
+      top: y,
       behavior: "smooth",
     });
 
-    if (window.innerWidth < 768) {
-      const body = document.querySelector("body");
+    intervalID = setInterval(function () {
```
erbridge, line 25:
```suggestion
    const intervalID = setInterval(function () {
```

Don't need to split it and you don't reassign it.

erbridge (commented):
Do we need this approach? Can we remove the focus immediately?

_stylesheets/styles.scss
```diff
@@ -39,6 +39,10 @@ h6 {
   font-family: Poppins, sans-serif;
   font-weight: 600;
   color: $navy;
+
+  &:focus {
+    outline: none;
+  }
```
erbridge, line 45: I think we do want some focus indication for those who prefer tab navigation.

pezholio:
No, I tried that, and setting the focus meant that the `scrollTo` animation wasn't happening

_javascript/lib/sidebarNavigation.js
```diff
@@ -13,19 +13,39 @@ export default (links) => {
...
+      const target = window.pageYOffset;
+      const min = target - 5;
+      const max = target + 5;
+
+      // Check if the Y Offset has got within 10 pixels of the target
+      // it's not always going to be exact, so make sure we're roughly
+      // there
+      if (y < max && y > min) {
```
erbridge, line 33:
```suggestion
      if (Math.abs(y - window.pageYOffset) < 5) {
```

(just a thought)

_javascript/lib/sidebarNavigation.js
```diff
@@ -13,19 +13,39 @@ export default (links) => {
...
+        document.activeElement.blur();
```
erbridge, line 34: What happens if the user shifts focus before the page is done scrolling?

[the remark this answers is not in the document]
pezholio: The clicked link, though I guess we have a reference to that, so might make more sense to blur that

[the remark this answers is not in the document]
erbridge: Yeah. what if the focus has changed.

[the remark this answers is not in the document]
pezholio: I did try that, and the animation happened too quick for me to be able to do it, but it's just as easy to do on the element, and is a good defence against this happening.

_javascript/lib/sidebarNavigation.js
pezholio: Good shout 👍

_javascript/lib/sidebarNavigation.js
```diff
@@ -10,22 +10,36 @@ export default (links) => {
...
 
-    if (window.innerWidth < 768) {
-      const body = document.querySelector("body");
+    const intervalID = setInterval(function () {
+      // Check if the Y Offset has got within 10 pixels of the target
+      // it's not always going to be exact, so make sure we're roughly
+      // there
+      if (Math.abs(y - window.pageYOffset) < 5) {
+        link.blur();
+        targetElement.setAttribute("tabindex", "-1");
```
erbridge, line 29: What's setting the `tabindex` do?

pezholio: By default, only links and form elements are focusable. The only way to make headings focusable is by setting a negative `tabindex`

erbridge: Could you add an explanation to the commit message maybe?

erbridge (approved):
If you could add the `tabindex` thing to the commit message, I think we're good to go.

_javascript/lib/sidebarNavigation.js
pezholio: I've added a comment to the code, just to make it clearer

PR #492, opened 2020-10-19
Link Playbook to Inclusion, diversity and equal opportunities policy

johnwaterworth:
Add a link from the current Diversity and inclusion statement to the new Inclusion, diversity and equal opportunities policy.


becshodges:
Hey John, I don't think I have the correct permissions to be able to do
this?

becshodges (approved):

PR #522, opened 2020-11-06
Update Ruby to 2.7.2

jacksonj04:
This resolves broken GitHub tests, caused by actions/setup-ruby@v1.1.2 not being able to locate version 2.7.1 ([example](https://github.com/dxw/playbook/runs/1345901628?check_suite_focus=true)).


Floppy (approved):
test pass, looks good to me. I had to do the same recently on some personal repos. Why 2.7.1 decided to start breaking I don't know.

PR #550, opened 2020-12-29
Add a guide to being a tech lead at dxw

erbridge:
This is an edit of @evilstreak's two Bikeshed posts on the topic:

- https://bikeshed.dxw.com/2020/03/20/being-a-tech-lead-at-dxw/
- https://bikeshed.dxw.com/2020/12/04/being-a-tech-lead-revisited/

### DIFF
--- a/guides/being-a-tech-lead-at-dxw.md
+++ b/guides/being-a-tech-lead-at-dxw.md
@@ -31,8 +31,8 @@ service needs to be:
 
 - maintainable with minimal effort
 - reliable, as there's no standing team to respond to issues
-- feature-complete enough that their users can do everything they need without
-  significant frustration
+- feature-complete enough and with a good enough user experience to meet their
+  service users' core needs
 
 Getting your software into that state is no mean feat, but knowing that's the
 end state can help your prioritisation from the start.
--- a/guides/being-a-tech-lead-at-dxw.md
+++ b/guides/being-a-tech-lead-at-dxw.md
@@ -136,15 +136,21 @@ A tech lead's job is to:
 
 - translate the problems to be solved into pieces of technical work for the team
   to pick up
+- advocate for building software that meets user needs
 - make sure the rest of the developers understand the domain and the work, and
   aren't blocked from doing it
 - encourage people to pair together on tricky work or to build team knowledge
+- teach other members of the team, directly and by example, encouraging them to
+  make sure their work is clear and maintainable
+- make sure that the team has the opportunity to learn
 - keep an eye on the development processes of the team, and address issues (e.g.
   stories continually taking much longer than expected, people being blocked by
   PRs not being reviewed, lots of time spent chasing down bugs)
 - be a technical representative for the team - taking (necessary) meetings and
   answering questions on behalf of the whole team, allowing them to focus on
   development work
+- be responsible for the team learning from failure and for incident reviews
+  happening when something goes wrong
 
 You will often, but not always, be the most experienced developer on the team.
 When you are you'll also be expected to make clear and informed architectural
--- a/guides/being-a-tech-lead-at-dxw.md
+++ b/guides/being-a-tech-lead-at-dxw.md
@@ -64,6 +64,15 @@ care about, and make sure your team is in a good position to satisfy them. Often
 those people you build relationships with will be happy to help if you involve
 them early enough!
 
+Sometimes the amount of governance we need to go through takes an unexpected
+amount of effort, particularly when the drive for governance comes from a
+different place to the drive for delivery. We do our best to meet all the
+governance needs but when that effort starts to risk delivery, we present those
+risks, along with alternatives, to the client. Ultimately, it's up to them what
+risk they're prepared to carry. We often have to do some amount of mediation in
+that case (usually biased towards delivery - it's what we're there for after
+all).
+
 ### ...for public services
 
 The public sector has a history of abdicating digital and technology decisions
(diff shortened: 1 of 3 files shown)


evilstreak (commented):
This looks good to me, but that's perhaps unsurprising given I drafted a lot of it!

I'd be interested to get reckons from a few of the folks who've done a bunch of this lately, like @tahb, @edavey and @mec.

guides/being-a-tech-lead-at-dxw.md
```diff
@@ -0,0 +1,149 @@
...
+better. When we leave, we need to leave them with working software and an
+in-house team that has the expertise to maintain and develop it.
+
+However, they might not be trying to build their own team. Although software is
+never done, we need to try our best to get there by the time we leave. Their
+service needs to be:
+
+- maintainable with minimal effort
+- reliable, as there's no standing team to respond to issues
+- feature-complete enough that their users can do everything they need without
```
tahb, line 34: I think "feature-complete" and "everything" could be interpreted as something that contradicts with "never done". This point sounds like it's trying to address user needs and usability, perhaps something like: "enabling users to deliver the minimum viable set of outcomes with a user experience that is good enough"?

guides/being-a-tech-lead-at-dxw.md
```diff
@@ -0,0 +1,149 @@
...
+Sometimes quite a lot of it.
+
+This extra work can feel really frustrating when you're trying to ship a
+service, but it's really important. Just as we wouldn't take on a service we had
+no hand in that doesn't meet any of our standards, we shouldn't expect our
+clients to. Part of your role is to build relationships with the technology
+folks in the client's organisation, find out which bits of governance we need to
+care about, and make sure your team is in a good position to satisfy them. Often
+those people you build relationships with will be happy to help if you involve
+them early enough!
```
tahb, line 65: Agree. Potentially someting more.

Sometimes the amount of governance and required processes we need to go through takes an unexpected amount of effort. Should we consider something further here that caveats this scenario? I believe will do our best to meet all governance but when that effort starts to risk delivery, we will present those risks along with alternatives to the client.

erbridge: There's a paragraph on this on 109, which maybe covers some of it?

guides/being-a-tech-lead-at-dxw.md
```diff
@@ -0,0 +1,149 @@
...
+doing this; get other folks in the team (like your delivery lead or technical
+architect) to help, and call on the wider dxw team when you need it.
+
+## What our tech leads do inside teams
+
+The role of a tech lead at dxw is to lead the technical work of the team. Given
+software is built by and for people, that means it's very much at the squishy
+intersection of technical expertise and people skills.
+
+A tech lead's job is to:
```
tahb, line 126: Some further ideas:

- advocate building software that meets user needs
- teach others directly and by example
- be responsible for the team learning from failure, ensure incident reviews happen and that the team has the opportunity to learn

edavey: @tahb good point about emphasising the "user-driven" viewpoint. I've found it's often helpful to bring teams' focus back to the user benefit we're trying to deliver.

Perhaps there could be a bullet point about 'quality control' which might follow the "teach others directly and by example" item e.g.

- encourage team members to go the extra 'yard' to make their work clear and maintainable

guides/being-a-tech-lead-at-dxw.md
evilstreak: The point of governance coming into conflict with delivery is a useful one. Particularly when the drive for governance and the drive for delivery come from different parts of our client organisation. In that situation we often have to do some amount of mediation (normally with a bias towards delivery, because that's why we're getting paid).

guides/being-a-tech-lead-at-dxw.md
evilstreak: Good point. Perhaps better phrased as "feature-complete enough and with a good enough user experience to meet their users' core needs"?

erbridge: Using Dom's suggested wording here.

guides/being-a-tech-lead-at-dxw.md
erbridge: Adding this paragraph immediately following, which I think covers it:

> Sometimes the amount of governance we need to go through takes an unexpected
> amount of effort, particularly when the drive for governance comes from a
> different place to the drive for delivery. We do our best to meet all the
> governance needs but when that effort starts to risk delivery, we present those
> risks, along with alternatives, to the client. Ultimately, it's up to them what
> risk they're prepared to carry. We often have to do some amount of mediation in
> that case (usually biased towards delivery - it's what we're there for after
> all).

guides/being-a-tech-lead-at-dxw.md
erbridge: Adding some bullets to cover these points.

erbridge:
Care for another look @edavey @tahb @evilstreak?

evilstreak (approved):
Changes look good 👍

guides/being-a-tech-lead-at-dxw.md
tahb: LGTM 👍

tahb (approved):

Floppy (approved):

Issue #412, opened 2020-07-24
Support sections need review


CristinaRO:
There are several places in the playbook where we talk about handling support. Some of them need specific changes, others need reviewing for redundancy or contradictions.
  * Team changes and handovers guide
    https://playbook.dxw.com/guides/team-changes-and-handovers.html
    The Rota spreadsheet is not the source of truth, it should mention OpsGenie as the primary source
  * Support helpdesk section
    https://playbook.dxw.com/#support-helpdesk
  * Support and on call guide
        https://playbook.dxw.com/guides/support-and-on-call#being-on-first-line-support

erbridge:
I do wonder if we should strip it out and rewrite it from the current context.

erbridge:
https://bikeshed.dxw.com/2020/12/11/changes-to-the-developer-support-rota/ is relevant to the team changes guide.

PR #557, opened 2020-12-29
Update the support documentation to better reflect reality

erbridge:
Support has changed a lot since the days we only supported WordPress sites, and our practices have changed with it.

Related to #412


erbridge:
It would be good to review this with an eye for things we don't do currently that we don't think we want to do.

rjw1 (approved):
looks good to me. its certainly what we aim for.

PR #564, opened 2021-01-04
Install Vale with Hombrew instead of go get

### LINKED ISSUE #563: Install Vale differently
CI seems to have stopped working for installing Vale. The docs suggest a different method:

https://docs.errata.ai/vale/install#using-vale-with-a-continuous-integration-ci-service

erbridge:
Vale is no longer installing correctly on CI. We don't use it there, and the Vale documentation says to install from Homebrew. This removes our dependence on Go and therefore on goenv.

Fixes #563


lozette (approved):
This all WFM!

PR #566, opened 2021-01-05
Support part-time working spread over longer than a fortnight 

erbridge:
This follows a conversation between the line managers about rolling out a pattern that has, so far, been a special case for some people at dxw, but could apply to anyone.


evilstreak (approved):
Looks good 👍 

Some of the existing language in this section could be written more plainly, but that's definitely something for a separate change.

PR #571, opened 2021-01-11
Fix Vale configuration issues

jdkato:
Hi,

I saw the commit linked from https://github.com/errata-ai/vale-action/issues/29 and I thought I'd try to help. 

It appears that this repo has a few configuration errors that were not caught when Vale was first implemented here (newer versions are capable of catching many more errors).

The last commit replaces the need to have `proselint` and `write-good` as submodules, which won't work as currently implemented. See https://github.com/errata-ai/vale/issues/273 for more information.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -9,6 +9,11 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - uses: errata-ai/vale-action@v1.3.0
+      - uses: errata-ai/vale-action@master
+        with:
+          styles: |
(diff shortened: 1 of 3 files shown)


Floppy:
Thanks @jdkato this looks good, thankyou!

Floppy:
Currently failing because of a lack of token in the checkout action - I think we might have the `env` section in the wrong place.

Floppy:
Actually I think it might be a permission thing. We can merge this and then check it works on master (seeing as it's broken on master anyway, we can't make it worse)

Floppy (approved):

PR #576, opened 2021-01-14
Do not use the dxw style on tone-of-voice.md

lozette:
tone-of-voice.md contains several "incorrect" examples. The linter is treating
these as errors (correctly) but we want them to be present to show what not
to write. Ignore this file with the dxw style to mitigate these errors.


erbridge:
If you're in the Vale config, I saw an option you can set to only run the GitHub action on changed files or content. Worth investigating.

PR #577, opened 2021-01-15
Replace submodules with direct downloads

erbridge:
The git repos have the wrong structure to be submoduled and used by Vale, so we download the latest version instead.


Floppy (approved):
This looks good to me! And it works on CI, though the status is red because vale is working :D

PR #585, opened 2021-01-19
Fix prose linting on existing playbook content

lozette:
Various fixes and linter adjustments to get the prose linting tests passing on the playbook.

- Do not use the dxw style on `tone-of-voice.md`, as this file contains deliberate mistakes to show what not to do
- Reduce write-good.ThereIs and write-good.So to warnings. These are grammatical undesirables but we use them a lot in the playbook
- Replace instances of '...' with the appropriate symbol. I did not want to change proselint.Typography to a warning, because this checker could pick up other more important errors down the line
- Reduce proselint.Cliches to a warning, we do use a few cliches but I think they are appropriate for our tone/content
- Reduce proselint.DateSpacing and proselint.DateCase to a suggestion, because our date formatting is fine for the playbook
- Reduce proselint.Skunked, proselint.Very and proselint.But to suggestions. These aren't errors in write-good, so this brings both linters in line with each other
- Reduce proselint.RASSyndrome to a warning. Acronyms that repeat themselves (e.g. "ATM Machine") are incorrect but also widely understood
- Change two instances of "TODO" to the more natural "To be added".

### DIFF
--- a/.vale.ini
+++ b/.vale.ini
@@ -16,3 +16,5 @@ proselint.GenderBias = YES
 write-good.So = warning
 write-good.ThereIs = warning
 proselint.Cliches = warning
+proselint.DateSpacing = suggestion
+proselint.DateCase = suggestion
--- a/.vale.ini
+++ b/.vale.ini
@@ -19,3 +19,4 @@ proselint.Cliches = warning
 proselint.DateSpacing = suggestion
 proselint.DateCase = suggestion
 proselint.Skunked = suggestion
+proselint.Very = suggestion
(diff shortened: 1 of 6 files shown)


erbridge (approved):
We should probably subdivide sections further to reduce the annotation noise when making changes.

Should probably get @duffylouise to have a look before merging (as the owner of the style guide).

duffylouise:
Thanks for checking - this all sounds good to me

erbridge:
The suggestion to subdivide wasn't intended to be part of this commit, but a note for the future.

PR #610, opened 2021-02-12
Rename to main

CristinaRO:
Update workflow and contributing guide to use `main` as the name of the main branch.

Presumably renaming the branch should happen roughly in the same time?


erbridge (approved):

erbridge:
Rename then merge?

PR #648, opened 2021-03-12
Install the same bundler version as in the Gemfile

CristinaRO:


CristinaRO:
Nope, this didn't do it.

PR #649, opened 2021-03-12
Update vulnerable npm packages

CristinaRO:
This is the result of running `npm audit fix`.


erbridge:
We only use Docsify for the CSS now, and it's pinned in the Dependabot config. We shouldn't update it in case we break the styles, and the CSS isn't affected by the vulnerability.

PR #656, opened 2021-03-19
Don't autocommit fixes on PRs from forks

lozette:
When a PR is opened from a fork (most commonly from Dependabot) any secrets
in the main git repo will not be available to the forked branch.

See https://github.blog/changelog/2021-02-19-github-actions-workflows-triggered-by-dependabot-prs-will-run-with-read-only-permissions/

Unfortunately this means this `test` Github action will always fail on Dependabot PRs because it tries to
autocommit fixes, and the secret required to autocommit fixes is not available
to the fork.

We don't think any PRs from Dependabot will require autocommitted linting fixes,
so it is safe to skip this action for PRs opened from forks.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -54,6 +54,7 @@ jobs:
 
       - uses: stefanzweifel/git-auto-commit-action@v4.8.0
         if:
+          github.event.pull_request.head.repo.full_name == github.repository &&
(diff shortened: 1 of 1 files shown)


erbridge:
This disables the test entirely. Can we instead, disable the autocommit instead?

erbridge (approved):

PR #691, opened 2021-04-27
Update hiring-new-people.md

becshodges:


johnwaterworth (changes requested):
Calling the additional interview a 'work day' doesn't make much sense given these changes. Maybe just call it a second interview.

And it can't be 'exactly the same' as the first interview.

You say that it has different team members and questions. How else might it be different? Is it more in depth for example?

And is it always at the start of the work day?

johnwaterworth (approved):

PR #702, opened 2021-05-11
Experimental: vale settings for linting only modified files

edavey:
**NOT TO BE MERGED, JUST A TEST**

Experimental: are we getting stuck on 'only modified'?

I'm wondering whether in the case of a PR which contains no
lintable changes, Vale hangs because it's waiting to hear
back from its 2 linters?

Perhaps this `onlyAnnotateModifiedLines` option is an
alternative if our intention is to limit feedback to the
content in the current PR?

i.e. everything will be linted on every PR but warnings and
errors will only be reported where they relate to changes in
the current PR.

Ref: https://github.com/dxw/playbook/pull/583/files

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -11,7 +11,6 @@ jobs:
 
       - uses: errata-ai/vale-action@master
         with:
-          onlyAnnotateModifiedLines: true
           styles: |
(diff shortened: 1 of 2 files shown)


edavey:
this has been resolved in https://github.com/dxw/playbook/pull/708, thanks Leeky!

PR #709, opened 2021-05-12
Move new Qualifications section down a level

johnwaterworth:
Support for professional qualifications should be at the same level as Personal learning and development allowance


edavey:
Looks good to me. 

I notice that the commit also changes the "Directors & heads-of" heading to a non-ampersand format (but doesn't changed the level of that heading)

edavey (approved):

becshodges (approved):

PR #776, opened 2021-08-12
Update lockfile version, Node and path-parse package

Gweaton:
This was motivated by a vulnerability in path-parse, and updating the package with my version of npm changed the lockfile version.

This should at least make it consistent for people running the latest npm version.


Gweaton:
Github actions has undone most of this - I think it will be resolved by automated dependabots anyway

PR #826, opened 2021-10-01
L and d tweaks

RobjS:
Tweaks to the L&D guide based on feedback from Vee.


vanessa-w (approved):
Thanks very much Rob - lgtm

PR #844, opened 2021-10-24
Add section on sharing research

johnwaterworth:
Say which slide templates to use, how to present activities and findings


erbridge (approved):
Approving from an "unblocking, it looks fine" perspective, not an "I approve this content" perspective.

PR #862, opened 2021-11-16
Link to Google Drive article on restricting access

johnwaterworth:


jdudley1123 (approved):
I wonder whether it would be worth having a separate section about Google Drive access that can then be linked to? It's something that is applicable beyond just user research, and that way we avoid having the same Google Help Center link twice.

johnwaterworth:
That's a good thought. We're still trying to figure out how best to use the different Google Workspace apps for these things - eg mail, calendar and drive all have different ways of labelling and controlling access to things. Once things are clearer, pulling that together would be useful.

PR #864, opened 2021-11-18
Update helper checklist link

jdudley1123:
The link was pointing to an old checklist that has since been superseded.


johnwaterworth (approved):
Link looks good. Could remove the /edit, but not a big deal.

PR #880, opened 2021-12-07
Add accessibility statement to who we are section

calumryan:
I'm proposing we add an overall statement about accessibility, not specific to just our website's level of accessibility.

This is a work in progress PR that I'd encourage others in different professions of dxw to add to with their own guides on how to approach making our ways of working accessible for all.

### DIFF
--- a/_includes/sections/who-we-are.md
+++ b/_includes/sections/who-we-are.md
@@ -164,27 +164,17 @@ Our approach is set out in our
 
 Thank you for being you - it makes us better.
 
-#### Accessibility statement
+#### Make everything we do accessible
 
-At dxw we want to make sure that everyone can access digital services including
-our own websites, communications and colleague resources.
+At dxw we want to make sure that everyone can access the digital services we
+create, including our own websites, communications and colleague resources.
 
-To create a consistent approach to digital accessibility we set out individual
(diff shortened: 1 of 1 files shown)


johnwaterworth (changes requested):
Completely agree with adding a specific statement about accessibility to our principles.

Can we give this a title more like the other principles: Start with people and their needs, Keep an agile mindset, Work at a sustainable pace, Reduce our impact on the environment?

Maybe:
Make everything we do accessible

johnwaterworth (commented):
Our statements of principle should be quite punchy.

Maybe we can simplify this to:

At dxw we want to make sure that everyone can access the digital services we create, including our own websites, communications and colleague resources.

We follow the guidance in the [GOV.UK Service manual](https://www.gov.uk/service-manual/helping-people-to-use-your-service/making-your-service-accessible-an-introduction) to help us to meet government accessibility requirements.

And to create a consistent approach to digital accessibility, our Playbook has [additional guidance for our different professions](/guides/web-accessibility).

johnwaterworth (changes requested):
I don't think the point about the accessibility statement for our website is relevant here.

johnwaterworth (approved):

PR #908, opened 2022-01-26
Rewrite the development workflow guide

erbridge:
Our existing guide was falling out of date with how we worked and was
very prescriptive and low level. This new guide zooms out a bit and
allows for more flexibility, leaning on providing guidance instead. We
lose some of the more step-by-step instructional content, which added a
lot of unnecessary bulk - new starters will have team members around to
show them how to do things.


rjw1 (approved):
I think these are a good improvement on what we had and will help us articulate our development practices to clients and other interested parties.

PR #914, opened 2022-02-04
Add Pay Policy

dragon-dxw:
Add updates to the pay policy as requesed by Vee.


vanessa-w (commented):
This is fine to add - there's some duplication between this policy and the Cost of Living section but as I can't make changes to this policy and there are things in the existing section not covered, I think we have to live with this until we review the policy again.

jacksonj04 (approved):
Assuming we get a 👍🏼  on the substantiative content, I can't see any glaring formatting/typo errors or things which just don't parse good.

PR #915, opened 2022-02-04
run 'bundle update'

dragon-dxw:


CristinaRO (approved):
In addition to the test suite passing, I've also run this locally and confirmed that the site still builds and functions as expected.

PR #933, opened 2022-03-15
Jekyll Refresh

james:
The existing site had 7 years of morphing from a markdown file to a docsify site and then to a jekyll site. This had accrued a lot of technical debt, and unconventional build processes.

We want to be able to enable netlify CMS, which meant dramatically simplifying the site.

The content in the commit is the same as before, but the wrapping static site generation has changed dramatically.


rjw1 (approved):
looks good

PR #935, opened 2022-03-15
Jekyll refresh continued

james:
Embarrassingly I failed to copy over the latest commits to the jekyll refresh in my last commit. This is another bulk of changes.


james:
Once this is merged, I'm happy for the DNS to be switched over.

It would also be good to remove the broken github workflow runs, which I don't think I have access to do.

erbridge (approved):

PR #969, opened 2022-08-19
Add scripts for setting up and running the application locally

Gweaton:
## Changes in this PR

- Adds Ruby installation tool-agnostic (as not everyone uses the same thing) scripts to rule them all: `script/setup`, `script/bootstrap` and `script/server`.
- Documents how to run the app locally in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,35 @@ The Playbook documents how the Playbook gets changed. See the
 [This Playbook](https://playbook.dxw.com/#this-playbook) section for more
 information.
 
+## Development guide
+
+### Setup
+
+Run the setup command. This installs the dependencies required for running the
+project.
+
+```
+script/setup
+```
+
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -15,6 +15,44 @@ The Playbook documents how the Playbook gets changed. See the
 [This Playbook](https://playbook.dxw.com/#this-playbook) section for more
 information.
 
+## Development guide
+
+### Dependencies
+
+These dependencies aren't installed by a setup script as people have their own
```
rjw1, line 22: I do wonder if we should be more prescriptive about this. If people dont use rbenv they are on their own for instance.

Gweaton: Have updated this after our discussion on Slack, as we tend to install `rbenv` etc here.

erbridge (approved):

PR #971, opened 2022-09-02
Clarify that some advice does not apply to GovPress

snim2:
Some hiring and progression information in both the playbook and the guides only applies to the Technology Team. This commit clarifies that the advice here does not apply to GovPress developers.

### DIFF
--- a/_guides/tech-team-junior-developer-development-plan.md
+++ b/_guides/tech-team-junior-developer-development-plan.md
@@ -1,6 +1,9 @@
 ---
-title: Junior developer development plan
+title: Technology team junior developer development plan
 ---
+
+_Please note that this document does not apply to the GovPress team._
+
 ## High level goals
 
 Before a junior developer can be promoted, they need to be:
--- a/_guides/tech-team-junior-developer-development-plan.md
+++ b/_guides/tech-team-junior-developer-development-plan.md
@@ -1,8 +1,8 @@
 ---
-title: Technology team junior developer development plan
+title: Junior developer development plan (Delivery+)
 ---
 
-_Please note that this document does not apply to the GovPress team._
+**Please note that this guide applies to the Delivery+ team only.**
 
 ## High level goals
 
--- a/_guides/tech-team-promotion-process.md
+++ b/_guides/tech-team-promotion-process.md
@@ -6,6 +6,8 @@ only. For gaps we identify in the team, one-off roles, or more senior roles
 (SFIA 6-7), we use a different process, often involving internal or external
 recruitment.
 
+_Please note that this document does not apply to the GovPress team._
+
 The process is based on
 [our progression framework](https://dxw.progressionapp.com/technology) and
 designed to help make fair decisions about promotions that reduce the impact any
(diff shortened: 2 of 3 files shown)


_guides/tech-team-junior-developer-development-plan.md
```diff
@@ -1,6 +1,9 @@
 ---
-title: Junior developer development plan
+title: Technology team junior developer development plan
```
erbridge, line 2:
```suggestion
title: Junior developer development plan (Delivery+)
```

_guides/tech-team-junior-developer-development-plan.md
```diff
@@ -1,6 +1,9 @@
...
 ---
+
+_Please note that this document does not apply to the GovPress team._
+
```
erbridge, line 6:
```suggestion

**Please note that this document applies to the Delivery+ team only.**

```

_guides/tech-team-promotion-process.md
```diff
@@ -6,6 +6,8 @@ only. For gaps we identify in the team, one-off roles, or more senior roles
 (SFIA 6-7), we use a different process, often involving internal or external
 recruitment.
 
+_Please note that this document does not apply to the GovPress team._
```
erbridge, line 9:
```suggestion
**Please note that this document applies to the Delivery+ team only.**
```

index.markdown
```diff
@@ -3263,7 +3263,9 @@ Designers at dxw follow these principles:
     needs their attention, and gives everyone in the team context for the work
     being done and decisions being made.
 
-#### Career progression for developers
+#### Career progression for developers in the technology team
+
+_Please note that this subsection does not apply to the GovPress team._
```
erbridge, line 3268:
```suggestion
#### Career progression for technologists in Delivery+
```

erbridge (commented):
Thanks for spotting this. I think "technology team" is also misleading, so I've put a bunch of suggestions in to change this to talk about "Delivery+", which is the real differentiator from "GovPress".

index.markdown
erbridge: It's probably worth saying that I'm not sure how much, if any of these sections apply to GovPress at all. We started to think about GovPress team principles a couple of years ago to go in alongside this stuff. I think we could probably do with a full review of the playbook to disambiguate D+ from GovPress everywhere.

erbridge (approved):

index.markdown
snim2: Also, I don't think we have the term Delivery+ on the company website, except in the Blog. I'm not sure whether this is something that needs changing, but I found it a little confusing.

snim2:
So, we need to decide whether to:

1. Tidy and merge this now, and raise a new issue to deal with separating out GovPress and Delivery+ throughout the Playbook; or
1. Wait until we have some synchronous time available to add all the necessary changes in this PR.

At a high level, what we need to do (here or in another PR) is:

1. Create a new **profession** for GovPress in the [section on our professions](https://playbook.dxw.com/#our-professions), maybe alongside **Strategy**
1. Copy over the existing development content and modify that for GovPress
1. Clarify that the original development content applies to Delivery+
1. `grep` through the rest of the Playbook and Guides to check we haven't missed anything

snim2:
Note about file changes: 

* If we just rename a file, then external links to it will break
* We need to re-create the deleted file, and add a link to GP/D+ versions of the document
* That file should be a hidden guide, and there might need to be some logic added to deal with that

PR #973, opened 2022-10-03
Restructure the playbook

erbridge:
This is a large change to break the playbook down into smaller sections. As such it contains a lot of changes. The best way to review is to look at the preview alongside the existing playbook and explore. Sorry!

The core changes:

- Drop the concept of "guides" in favour of nested content
- Carve the index page up into many smaller pieces and interleave with what used to be guides
- Demote the importance of the page outline sidebar and hide it by default on all screen widths
- Introduce a new "related content" sidebar to allow navigating up and down the tree
- Deep link the contribution links to edit the page you're looking at


jdudley1123:
I'm very much in favour of ditching the guides and nesting the content, so I think this is a step in the right direction, but I really struggle to navigate in this version:
- The navigation text is very large and all in bold, with lots of entries going across two lines (this is a holdover from the current design, but made more obvious by the larger text and higher contrast colours)
- Clicking on a nav link sometimes takes you into a page that has no children, so you end up with a massive empty navbar
- The purpose of the hamburger menu is unclear to me - a hamburger menu generally conceals site navigation, but in this case I think it's in-page navigation, and generally only has a few headings that don't seem worth concealing in a whole navbar

I think navigation similar to [Read the Docs](https://docs.readthedocs.io/en/stable/tutorial/) with a single nav bar and expandable nesting would be easier to use.

erbridge:
@jdudley1123 I definitely take your point on the colours and sizing. I've tweaked the sidebars to be less dominant.

On the nav structure, if I were to make one change, it would be to drop the (in-page) outline sidebar entirely. That might be a sensible next step, but I was trying to be additive with this change rather than subtractive to avoid too many "but my favourite feature is missing" comments holding it up. The outline sidebar is self-contained, so removing it should be fairly straightforwards (remove the `include` call for it in the default layout and then clean up the now unused code).

On the empty related content sidebar, we could only show it when there are children, or (as I had originally intended) add cross-linking to other related pages into that view (via eg a frontmatter list called `related_pages` or something), so it's more than just child links, but is more of an index of related content. I found that making it come and go as you navigate the site made it feel a bit jarring as the whole view would re-layout, so opted to keep it even when empty for now.

All of the content is the same as what already exists, and another next step might be to fill out of some the index pages (which should render as an index of links, not as blank - did you find a counter example?). In the interests of incremental improvements (and because I didn't feel it was right for me to rewrite any of the playbook content on my way out of the company), I've left content improvements as an exercise for the future.

Overall, from my perspective this is a step in the right direction. There might be further improvements worth making, but these changes make the playbook overall better, in my opinion, and sets it up for future improvements. My ultimate guiding question for reviewers is "is this better than what exists already?". If this answer is "yes", I suggest we go with this and then improve it from there instead of holding this back until it's "perfect".

src/tech/how-we-do-development.md
```diff
@@ -1,13 +1,17 @@
 ---
 title: How we do development
+permalink: /:path/:basename/
+redirect_from:
+  - /guides/how-we-do-development
```
snim2, line 5: Just a note that once this has been merged, I'll either need to rebase and rewrite #971, or close it if this restructuring can resolve the original concern (that it wasn't clear which advice applied to GP and which to D+).

README.md
```diff
@@ -44,6 +44,15 @@ script/server
 
 This runs the server on localhost:4000.
 
+### Checking the HTML
+
+To proof the HTML for things like dead links, run:
+
+```
+bundle exec jekyll build
```
snim2, line 52: This could be in `/script/test`

src/tech/index.md
```diff
@@ -169,10 +169,9 @@ where we describe what changed and why we changed it. We keep track of any
 future changes we may want to make or think about in the
 [backlog](https://docs.google.com/document/d/16pZuxt_CWp2NFaYaqBIZvuHmPeNWcqKycLkx5ciErbw/edit).
 
-We have a [documented promotion process](/guides/tech-team-promotion-process)
-based on that framework. We designed it to help make fair decisions about
-promotions that reduce the impact any one person can have on the decision as
-much as possible.
+We have a [documented promotion process]navipromotion-process) based on that
```
snim2, line 173: This link is broken (maybe we could use the [`mdl`](https://github.com/markdownlint/markdownlint) gem to lint Markdown? It's quite easy to [configure](https://gitlab.com/beautifulcanoe/peopleops/docs.beautifulcanoe.com/-/blob/main/.mdl.rb)

src/user-research/how-we-do-user-research.md
```diff
@@ -164,9 +167,8 @@ ethical, inclusive and sustainable ways.
 We want to involve our colleagues and other stakeholders in creating research
 plans, so they understand and can contribute to research activities.
 
-And we want to create plans that allow us to
-[learn, share and adapt](https://playbook.dxw.com/#user-research-principles) as
-we go.
+And we want to create plans that allow us to [learn, share and adapt](..) as we
```
snim2, line 170: Broken link?

src/contributing/managing-the-playbook.md
```diff
@@ -4,6 +4,8 @@ permalink: /:path/:basename/
 redirect_from:
   - /guides/managing
 ---
+**Needs review**
```
snim2, line 7: Is there a Trello card or an issue for this? It probably needs someone to coordinate a discussion?

snim2 (changes requested):
This looks great, but there are a few small issues I found that could do with fixing before this is merged (e.g. broken links), and a couple more that you might want to push out to separate issues / PRs.

src/user-research/how-we-do-user-research.md
erbridge: That's a fully functioning relative link, isn't it?

src/tech/index.md
erbridge: We dropped all linting because it gets in the way of non-technologist editors doing their thing. This is a typo, though.

README.md
erbridge: It could but it has 10 errors due to private links, so I didn't want to encode it as something with pass / fail semantics.

src/tech/how-we-do-development.md
erbridge: I suggest the former after a review.

src/contributing/managing-the-playbook.md
erbridge: Nope. Please do start that conversation. As I said above, I'm not trying to tackle any of the content here.

src/user-research/how-we-do-user-research.md
snim2: You're right!

src/tech/index.md
erbridge: Fixed the typo.

README.md
erbridge: Docs are here if our future selves want to make those 10 errors go away, so we can put this in a test: https://github.com/gjtorikian/html-proofer

snim2: OK, I found I needed `--check-internal-hash=false` to get a similar result, but the link to https://www.gov.uk/government/publications/guide-to-the-general-data-protection-regulation maybe does need changing?

snim2: ^^^ Note for me: this needs a new issue

snim2 (approved):

snim2:
@erbridge Thanks, I've marked this as approved and I'll raise issues for the things we need to follow up in a bit

james:
Fantastic work F, this is a big bit of work to resolve a structural problem that the playbook had, putting us on a good footing to continue iteratively improving the playbook from here.

james (approved):

src/_sass/_related.scss
```diff
@@ -0,0 +1,77 @@
+@use "./breakpoints";
+@use "./colours";
+@use "./typography";
+
+.related {
+  width: 100%;
+  padding: 1em;
+  background-color: colours.$accent-a;
```
yndajas, line 8: Is this an intentional background colour change? Previously it was `colours.$cream`

yndajas: Is it a hangover from splitting out the main navigation and page outline sidebars?

PR #979, opened 2022-10-06
Update link to wpscan in advisories page

### LINKED ISSUE #976: Update link to wpscan
In the section about WP plugin advisories, the link to https://wpvulndb.com/ now resolves to: https://wpscan.com/ and should probably be replaced.

This is currently waiting on #973 to be merged.

snim2:
The link to wpvulndb.com now resolves to wpscan, so it seems sensible just to link directly to the new URL.

Fixes #976

### DIFF
--- a/src/govpress-unit/plugin-advisories.md
+++ b/src/govpress-unit/plugin-advisories.md
@@ -26,7 +26,7 @@ the
 ## Check for existing advisories
 
 Pluginscan includes a report on vulnerabilities, based on the API exposed by the
-[WPScan Vulnerability Database](https://wpvulndb.com/)
+[WPScan Vulnerability Database](https://wpscan.com/)
 
 It's important that we avoid publishing advisories that have already been
 reported by someone else.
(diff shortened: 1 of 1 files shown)


RobjS:
@snim2 I don't think is actually true anymore. The WPScan API has long been behind a paywall, so the requests just get blocked.

We recently migrated GovPress Tools migrated to use [Patchstack](https://patchstack.com/) for its vulnerability data. That's also behind a paywall, but a much cheaper one than WPScan! We should either update pluginscan to use PatchStack (if you provide a valid API key), or not do those checks at all, as there's no longer any well-maintained free API for WordPress plugin vulnerabilities.

snim2:
@RobjS so, it sounds like we need to re-write this page a bit?

RobjS:
@snim2 I think for now we just need to say that you should check the [PatchStack vulnerabilities list](https://patchstack.com/database/) to ensure the vulnerability hasn't already been reported. But in the longer term we should remove the WPScan functionality for the pluginscan tool, or replace it with the PatchStack equivalent.

RobjS (approved):
👍

PR #983, opened 2022-10-12
Fix typographical errors and inconsistencies

yndajas:
Fixes a bunch of spelling mistakes and makes capitalisation usage more consistent


snim2 (approved):
@yndajas Thanks for doing this!

PR #989, opened 2022-10-22
Add search: iterations

yndajas:
This ticks off most of the to do lists in #984 as well as in a #dxw-playbook discussion on Slack, a few extra things George and I discussed and a few more than cropped up in the process

Outstanding:
- Analytics
- Delaying page load? I wonder whether we should think about delaying page load until some of the processing is done (if this doesn't go too far against progressive enhancement), since currently the search input will noticeabley display placeholder text before being updated to the search term
- Tidy up the script code - made a few quick comments on #984 but I think it would be good to look through the whole of the `search.js` script to improve the style/clarity/readability. A few small improvements jump out across the file

Bunch of commits here to break everything down and make it easier to review, but we could consolidate a little later if preferred

## Screenshot

image

### DIFF
--- a/src/_includes/navbar.html
+++ b/src/_includes/navbar.html
@@ -6,6 +6,9 @@
       >
     </li>
     <li>
+      {% include search_form.html %}
+    </li>
+    <li>
       <a href="https://www.dxw.com">dxw.com</a>
     </li>
   </ul>
--- a/src/_includes/navbar.html
+++ b/src/_includes/navbar.html
@@ -5,11 +5,14 @@
         ><img src="/assets/d-marque.png" role="presentation" alt="" /> Home</a
       >
     </li>
-    <li>
+    <li class="navbar__search--inline">
       {% include search_form.html %}
     </li>
     <li>
       <a href="https://www.dxw.com">dxw.com</a>
     </li>
   </ul>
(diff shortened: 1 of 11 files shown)


Gweaton (approved):
This is incredible work, well done! I've tested it out locally and it looks great. I think we can merge it into my PR and fixup a bit. 

There's a bit of weirdness going on when you search for more than one term:

image

Is it because it matches the title, do you think?

For multiple words, we may want to tweak the rankings for if a match is found in a title vs if the multiple words are consecutive. I think it'll take a bit of investigation, maybe we can do that together?

yndajas:
> There's a bit of weirdness going on when you search for more than one term:
> 
> Is it because it matches the title, do you think?
> 
> For multiple words, we may want to tweak the rankings for if a match is found in a title vs if the multiple words are consecutive. I think it'll take a bit of investigation, maybe we can do that together?

So I think Lunr is matching based on all the search terms semi-independently (possibly with added weight for when they appear in the order you provided them?), whereas our highlighting (taken from the guide?) is matching the exact term and nothing else. So yeah, you'll sometimes see zero matches and no highlighting because the match is not exactly the entire search phrase. One to think about, definitely!

PR #990, opened 2022-10-24
Add favicon based on d-marque asset

snim2:


src/_layouts/default.html
```diff
@@ -6,7 +6,7 @@
       name="viewport"
       content="width=device-width, initial-scale=1, shrink-to-fit=no"
     />
-
+    <link rel="shortcut icon" type="image/x-icon" href="/assets/favicon.ico?">
```
yndajas, line 9: Is the `?` in the `href` meant to be there?

yndajas (approved):
At last! 🎉

I wonder if it's worth adding a full favicon/PWA icon suite, e.g. the ones that are generated by [realfavicongenerator.net](https://realfavicongenerator.net)? That gives better support for different platforms and non-traditional browser usage, though that's likely an edge case for this repo

src/_layouts/default.html
snim2: No, good catch!

src/assets/site.webmanifest
```diff
@@ -0,0 +1,19 @@
...
+            "type": "image/png"
+        },
+        {
+            "src": "/android-chrome-512x512.png",
+            "sizes": "512x512",
+            "type": "image/png"
+        }
+    ],
+    "theme_color": "#ffffff",
+    "background_color": "#ffffff",
```
yndajas, line 17: Likewise here

PR #993, opened 2022-11-04
Replace the d-marque with the full dxw logo

### LINKED ISSUE #992: Use full dxw logo
Suggested by Head of Design in a [Slack thread](https://dxw.slack.com/archives/CK5HUK6SE/p1667581515445719?thread_ts=1667574484.992329&cid=CK5HUK6SE) asking for feedback on the new search functionality - it's not within the scope of that piece of work, so I'm opening an independent issue for this

snim2:
Closes #992

I've used the `.png` version of this logo, sized to the height of the previous logo. A `.svg` logo is available and listed as "unsized" on the branding site, but it's viewport is large (`1200x416.5`).


yndajas:
That was quick! 🎉

I'm not sure about the "Home" text now that we have the full logo. The ascender in "d" throws the sense of visual alignment off a little for me - it feels like "Home" should be aligned with the ascender-less "xw" part, but it might also need a bit more margin to push it away from the logo

I wonder how useful it is though - the logo feels like a more obvious/conventional way to navigate home than the "D" marque, so I wonder if the explicit "Home" is now redundant? Whether it goes or not, it would be good to get an `alt` on the image. If it does go, do we need an accessible property on the link saying where it'll take users?

Maybe one to throw back to Gaz as the requester!

snim2:
> I wonder how useful it is though - the logo feels like a more obvious/conventional way to navigate home than the "D" marque, so I wonder if the explicit "Home" is now redundant? Whether it goes or not, it would be good to get an `alt` on the image. If it does go, do we need an accessible property on the link saying where it'll take users?
> 

I agree, better without the "Home" text, and I think for accessibility we probably need an `aria-label` on the link.

I'll push a change and ask people to review on Monday.

src/_includes/navbar.html
```diff
@@ -1,9 +1,9 @@
 <header class="navbar">
   <ul aria-label="Main navigation">
```
calumryan, line 2: The aria label isn't read out by assistive tech since this isn't a landmark, I would suggest wrapping in a navigation element, e.g. `<nav aria-label="Main">`

src/_includes/navbar.html
```diff
@@ -1,9 +1,9 @@
...
     <li>
-      <a href="/"
-        ><img src="/assets/d-marque.png" role="presentation" alt="" /> Home</a
-      >
+      <a href="/" aria-label="Homepage">
+          <img src="/assets/dxw-logo.png" aria-hidden="true" alt="" />
```
calumryan, line 5: There shouldn't be a need for `aria-hidden="true"` here since an empty alt property indicates the image is decorative and therefore hidden from the accessibility tree of elements needing to be read out.

src/_includes/navbar.html
snim2: Wrap the whole `<ul></ul>`?

calumryan: Yes similar to how it's done on dxw.com
```
<nav aria-label="Main">
<ul></ul>
</nav>
```

calumryan (approved):

yndajas:
The logo underline looks really nice! Small thing that's probably not noticeable without zooming in, but it looks like there's some space between the text and underline that isn't on the logo

image

image

yndajas (approved):
🎉

PR #998, opened 2022-11-14
Style the navbar as white on navy

### LINKED ISSUE #997: Change logo and navbar colour
Change the header colour to our navy blue and use one of the white logos.

snim2:
Closes #997


yndajas (approved):
Looks good! I'm guessing the colours in `browserconfig.xml` and `site.webmanifest` are still suitable?

snim2:
> Looks good! I'm guessing the colours in `browserconfig.xml` and `site.webmanifest` are still suitable?

Good question. The theme colour is ivory, which I think is probably fine. Somehow the tile colour is `#da532c` which was really never right, I think, so I'll change that to the navy.

PR #1001, opened 2022-11-15
FIX 404 errors on admin pages

### LINKED ISSUE #1000: Edit this page link does not show page content
This is likely the result of !973 but we've had a few reports of the `Edit this page` not showing the page content that the user wanted to edit.

snim2:
Closes #1000

### Notes on testing this PR

To see the admin pages working in the review app, it's probably easiest to go direct to a relevant URL, e.g.: https://deploy-preview-1001--laughing-payne-b9fbd2.netlify.app/admin/#/collections/home

The review app for an unrelated PR can be compared here: https://deploy-preview-998--laughing-payne-b9fbd2.netlify.app/admin/#/collections/home

### Notes on the changes here

When running netlify-cms locally on `main`, this bug could not be reproduced. On production 404 errors are being received from the GitHub API page, e.g. on [this admin page](https://playbook.dxw.com/admin/#/collections/product-management) the console shows a 404 from:

https://api.github.com/repos/dxw/playbook/git/trees/main:product-management

whereas this URL:

https://api.github.com/repos/dxw/playbook/git/trees/main:src/product-management

returns 200 OK. So it seems that the GitHub back-end needs paths relative to the repo root, so that it can correctly query the GitHub API.

The documentation isn't very clear, but this page:

https://www.netlifycms.org/docs/collection-types#file-collections

says:

"""
Each file has its own list of fields and a unique filepath specified in the file field (relative to the base of the repo). 
"""

There isn't an equivalent statement for folder collections.

This PR changes the file paths, and also updates the instructions for running the CMS features locally (where paths can be relative to the content folder).

### DIFF
--- a/src/admin/config.yml
+++ b/src/admin/config.yml
@@ -111,16 +111,6 @@ collections:
       - { label: Permalink, name: permalink, widget: hidden, default: /:path/:basename/ }
       - { label: Body, name: body, widget: markdown, required: true }
 
-  - name: finance
-    label: Finance
-    label_singular: Finance page
-    folder: src/finance
-    create: true
-    fields:
-      - { label: Title, name: title, widget: string, required: true }
-      - { label: Permalink, name: permalink, widget: hidden, default: /:path/:basename/ }
-      - { label: Body, name: body, widget: markdown, required: true }
-
   - name: leadership
     label: Leadership
     label_singular: Leadership page
--- a/src/admin/config.yml
+++ b/src/admin/config.yml
@@ -1,9 +1,16 @@
-# To run locally, replace the `backend` config with the lines below and run:
-#   cd src && npx netlify-cms-proxy-server
+# To run netlify-cms locally:
+#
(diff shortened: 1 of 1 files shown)


src/admin/config.yml
```diff
@@ -107,7 +114,7 @@ collections:
   - name: finance
     label: Finance
     label_singular: Finance page
-    folder: finance
+    folder: src/finance
     create: true
```
yndajas, line 118: This folder doesn't appear to exist anymore

yndajas (approved):
Looks good. There's a folder in there that doesn't exist (`finance`) and the `contributing` folder isn't included, but the latter might be on purpose?

snim2:
> Looks good. There's a folder in there that doesn't exist (`finance`) and the `contributing` folder isn't included, but the latter might be on purpose?

I think the latter probably is on purpose, but I'll push a commit to remove `finance`. I think this went into `commercial-ops`

Issue #1031, opened 2022-11-25
Integrate Standard linting into CI


yndajas:
#984 introduced [JavaScript Standard Style](https://standardjs.com) to the Playbook via a package. This enables linting of JavaScript files to ensure they conform to standard style conventions. If using the companion VSCode plugin or something similar, you can also set up automatic fixing when saving

Before #984, there were no JavaScript files in the repo, but it would be good to integrate Standard linting into the CI pipeline to provide assurance that all future JavaScript code (whether related to search or anything new) is conformant before being merged into `main`

snim2:
@yndajas Do you get warnings in the current `main`?

```shell
standard: Use JavaScript Standard Style (https://standardjs.com)
  .../playbook/src/assets/js/search.js:21:19: 'elasticlunr' is not defined. (no-undef)
  .../playbook/src/assets/js/search.js:26:28: 'lunr' is not defined. (no-undef)
```

yndajas:
I do see those warnings in VSCode - I wonder if it's because we're including `elasticlunr` via a script tag rather than an `import` statement. I'm not sure how we get `lunr` - it's set within `elasticlunr.min.js` but doesn't appear to be exported

Any ideas?

yndajas:
We could try installing via npm instead: https://github.com/weixsong/elasticlunr.js#7-use-elasticlunr-in-nodejs. Still not sure if that would resolve the undefined `lunr` though. We could always disable the warning on a per-line basis before integrating Standard into the CI pipeline

lookupdaily:
Hello. Have opened two PRs related to this, one for resolving these standard errors and one for integrating Standard into the pipeline. 

Have resolved the import error by adding a JavaScript compiler (babel) and webpack - which allows us to use file level imports - but let me know if you think this approach is not suitable.

**PRs**
#1082 
#1083 - linting check will fail until we resolve the first but just have this open for reference

PR #1055, opened 2023-01-11
Handle various redirects

### LINKED ISSUE #1032: Investigate #-based redirects
## Problem

In restructuring user research pages, @johnwaterworth sought to redirect from URLs with hash fragments, such as `/guides/how-we-do-user-research#how-we-protect-participant-privacy` or `/user-research/how-we-do-user-research#how-we-protect-participant-privacy`. Given that [JekyllRedirectFrom](https://github.com/jekyll/jekyll-redirect-from) works by creating a HTML file for each URL you want to redirect from (within which is a meta tag that tells the browser to redirect), and `#` is not a valid filename character, it doesn't seem capable of supporting this feature

This issue stems from the shift in #973 from 90+% of content existing on a single page to having many sections each with many pages. Before we relied heavily on hash fragments to direct people within a huge single page, so with the restructure we have a lot of external hash-reliant dead links. The examples from John suggest we may also want redirects from sections within pages from the old guides section, as well as from sections within pages in the new structure

## Solutions

We could do one of the following:
- add code to the 404 page that looks at `window.location.pathname` and `window.location.hash`, looks up some (JSON?) redirect table for a match and redirects if one is found
- manually create the redirect from pages (probably as HTML files, rather than Markdown), and within them include JavaScript to redirect based on hash fragments

While changes to redirects would hopefully be small and infrequent, both of these solutions feel a little too reliant on developer maintenence. Is there anything we can do with Netlify CMS to allow edits to a JSON file, so that we can enable anyone to easily open PRs to maintain redirects?

adamhughes86:
Adds the redirect.html as an override which is detailed in the [jekyll-redirect-to repo](https://github.com/jekyll/jekyll-redirect-from/blob/master/lib/jekyll-redirect-from/redirect.html). This has then been modified to grab the location hash and append it to the redirect url.

Tested locally using [https://playbook.dxw.com/guides/support-and-on-call.html#claiming-toil-for-out-of-hours-alerts](https://playbook.dxw.com/guides/support-and-on-call.html#claiming-toil-for-out-of-hours-alerts). Not sure if it covers all hash redirect issues but seems to improve some. 

Resolves #1032

### DIFF
--- a/src/_layouts/redirect.html
+++ b/src/_layouts/redirect.html
@@ -1,0 +1,11 @@
+<!DOCTYPE html>
+<html lang="en-US">
+  <meta charset="utf-8">
+  <title>Redirecting&hellip;</title>
+  <link rel="canonical" href="{{ page.redirect.to }}">
+  <script>location = "{{ page.redirect.to }}" + location.hash</script>
+  <meta http-equiv="refresh" content="0; url={{ page.redirect.to }}">
+  <meta name="robots" content="noindex">
+  <h1>Redirecting&hellip;</h1>
+  <a href="{{ page.redirect.to }}">Click here if you are not redirected.</a>
+</html>
--- a/src/_layouts/redirect.html
+++ b/src/_layouts/redirect.html
@@ -1,11 +1,35 @@
 <!DOCTYPE html>
-<html lang="en-US">
-  <meta charset="utf-8">
-  <title>Redirecting&hellip;</title>
-  <link rel="canonical" href="{{ page.redirect.to }}">
-  <script>location = "{{ page.redirect.to }}" + location.hash</script>
-  <meta http-equiv="refresh" content="0; url={{ page.redirect.to }}">
-  <meta name="robots" content="noindex">
-  <h1>Redirecting&hellip;</h1>
-  <a href="{{ page.redirect.to }}">Click here if you are not redirected.</a>
+<html lang="en-GB">
+  <head>
+    <meta charset="utf-8" />
+    <meta
+      name="viewport"
+      content="width=device-width, initial-scale=1, shrink-to-fit=no"
+    />
+    <link rel="apple-touch-icon" sizes="180x180" href="/assets/apple-touch-icon.png">
+    <link rel="icon" type="image/png" sizes="32x32" href="/assets/favicon-32x32.png">
+    <link rel="icon" type="image/png" sizes="16x16" href="/assets/favicon-16x16.png">
+    <link rel="manifest" href="/assets/site.webmanifest">
+    <link rel="mask-icon" href="/assets/safari-pinned-tab.svg" color="#5bbad5">
+    <meta name="msapplication-TileColor" content="#243746">
+    <meta name="theme-color" content="#f9f8f5">
+    <title>Redirecting&hellip;</title>
+    <link rel="canonical" href="{{ page.redirect.to }}">
+    <script>location = "{{ page.redirect.to }}" + location.hash</script>
+    <meta http-equiv="refresh" content="0; url={{ page.redirect.to }}">
+    <meta name="robots" content="noindex">
+
+    <link rel="stylesheet" href="/assets/main.css" />
+    <script defer data-domain="playbook.dxw.com" src="https://plausible.io/js/plausible.js"></script>
+  </head>
+  <body>
+    {% include navbar.html %}
+    <main>
+      {% include related.html %}
+      <article class="page-content">
+        <h1>Redirecting&hellip;</h1>
+        <a href="{{ page.redirect.to }}">Click here if you are not redirected.</a>
+      </article>
+    </main>
+  </body>
 </html>
--- a/src/assets/js/redirect.js
+++ b/src/assets/js/redirect.js
@@ -1,12 +1,12 @@
-var fullPath = location.pathname + location.hash
+const fullPath = location.pathname + location.hash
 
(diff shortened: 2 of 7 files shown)


snim2:
Thanks for this, @adamhughes86

snim2 (approved):

snim2:
@yndajas I'm not sure whether you want to leave that issue up for a bit longer, just to check that this covers everything?

src/_layouts/redirect.html
```diff
@@ -0,0 +1,11 @@
+<!DOCTYPE html>
+<html lang="en-US">
```
yndajas, line 2: Shall we keep this as `en-GB` in line with `default.html`?

src/_layouts/redirect.html
```diff
@@ -0,0 +1,11 @@
...
+  <meta charset="utf-8">
+  <title>Redirecting&hellip;</title>
+  <link rel="canonical" href="{{ page.redirect.to }}">
+  <script>location = "{{ page.redirect.to }}" + location.hash</script>
+  <meta http-equiv="refresh" content="0; url={{ page.redirect.to }}">
+  <meta name="robots" content="noindex">
+  <h1>Redirecting&hellip;</h1>
+  <a href="{{ page.redirect.to }}">Click here if you are not redirected.</a>
```
yndajas, line 10: Guess it doesn't matter, but is it worth including the conventional `head` and `body` tags?

yndajas (approved):
Nice! Couple of trivial comments, but this seems to work great

src/_layouts/redirect.html
adamhughes86: Yeah that makes sense. It's in our codebase now should so match the standards. I'll update this template based on the default.html

src/_layouts/redirect.html
adamhughes86: I didn't even notice they weren't in there. I'll update to match the standards in the default.html

adamhughes86:
@yndajas @snim2 

I've added more onto this pr related to the redirects. I believe this should cover all requirements now but it could do with a good review as I've not used netlify before.

- Created reusable redirect script and applied to homepage and 404 file.
- Created redirect manager for netlify.
- Added redirect file that netlify will append to with first couple of redirects

src/assets/js/redirect.js
```diff
@@ -0,0 +1,12 @@
...
+fetch('/redirects.json')
+  .then((response) => response.json())
+  .then((json) => {
+  var redirectArray = json.redirects
+  var redirectResult = redirectArray.find(item => item.from === fullPath);
+
+  if (redirectResult) {
+    location = redirectResult.to
+  }
+});
```
yndajas, line 12: Since #984 we've adopted [Standard](https://standardjs.com) on this repo. It's not yet integrated with the CI (see https://github.com/dxw/playbook/issues/1031), so non-conformant code won't produce errors, but it would be good to keep new JS code in line if possible. Using an auto-formatting extension in VSCode, it fixes this to the following to align with Standard:

```
const fullPath = location.pathname + location.hash

fetch('/redirects.json')
  .then((response) => response.json())
  .then((json) => {
    const redirectArray = json.redirects
    const redirectResult = redirectArray.find(item => item.from === fullPath)

    if (redirectResult) {
      location = redirectResult.to
    }
  })
```

Would you be able to update the code to match?

yndajas (approved):
This works great! Tested locally and the redirects worked as expected, and I was able to edit the list of redirects via the Netlify CMS 🎉

I've made one comment on the script style. A few of other things came to mind, but nothing that should be a blocker if you want to get this merged (can always write these up as separate issues for future work):
1. Could we update `CONTRIBUTING.md` to explain how redirects work? We now have two different ways of creating redirects:
    - using the `jekyll-redirect-from` gem, adding a `redirect_from` entry to a page's front matter - this works for redirecting one page to another, including keeping the same hash fragment thanks to your first commit
    - using the Netlify CMS redirect manager - this works on the homepage and when you get a 404, and is made to plug a specific gap in the above: redirecting from a section of an old page (via a hash fragment) to a new dedicated page
2. Would it be worth wholesale replacing `jekyll-redirect-from` with this new approach? Perhaps it's best to keep the gem, since it's easier to spot/understand a redirect in the front matter of a specific page than in a long list, and just use the redirect manager by exception
3. The 404 page shows briefly before redirects - is it possible to get this script to halt page load/render until complete, so we don't get the quick 404 flash? I can't remember how feasible that is in fairly vanilla JavaScript/HTML
4. A nice to have would be for the JSON and/or Netlify CMS to order the redirects alphabetically. If the list gets long, it might be hard to see if a redirect already exists (although if it does, you probably wouldn't be trying to add it). The list is sortable via drag and drop, but that's reliant on user intervention. As far as I can see, the list widget doesn't support this natively

src/assets/js/redirect.js
adamhughes86: Yeah no problem. I'll sort that now

adamhughes86: In reference to the points above:

1. I will add some details to the CONTRIBUTING.md file.
2. I think as long as it's documented we should keep both for now and only use the redirect manager if required. We could change this down the line if we think it's worth it.
3. I'll add an extra ticket. We could probably put a simple bit of JS in there to hide the 404 container until the redirect script is complete
4. That would be a good feature. I'll add it to the issues for now

Issue #1060, opened 2023-01-23
Add jekyll-sitemap


yndajas:
As suggested by @jacksonj04, add [jekyll-sitemap](https://github.com/jekyll/jekyll-sitemap) to the Playbook to automatically generate a sitemap

PR #1063, opened 2023-02-01
Add jekyll-sitemap plugin to automatically generate sitemap

lookupdaily:
Add [`jekyll-sitemap`](https://github.com/jekyll/jekyll-sitemap) to automatically generate a sitemap on build.

Have excluded '/admin' from the sitemap as this does not contain any searchable contents.
The sitemap is currently only listing the `<loc>` tag, as I believe we are not tracking the last modified date. 

Addresses issue #1060


yndajas (approved):
Looks good to me. I ran the server locally and the sitemap contains the pages we'd expect, with admin excluded, and it isn't showing up in search results

PR #1064, opened 2023-02-02
Clarify approach to code reviews

lookupdaily:
Previously the text on the process for merging a pull request referred to 2 reviewers including the author(s), which could be confusing. It was written in this way to accommodate a pair and mob situation-which is clarified int his update.

Have also reworded the next sentence slightly for readability, as there appeared to be a missing word.

### DIFF
--- a/src/tech/how-we-do-development.md
+++ b/src/tech/how-we-do-development.md
@@ -164,7 +164,7 @@ origin [BRANCHNAME]`.
 ### 8. Submit the changeset for code review
 
 At dxw, we never make changes to code running in production without the involvement of at least two people. In practice, this means one author and at least one approving reviewer, or authoring code as a pair or mob. We do this to make
-sure we're delivering high quality software and it's an opportunity to learn from each other.
+sure we're delivering high quality software and use it as an opportunity to learn from each other.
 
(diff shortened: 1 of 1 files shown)


snim2 (approved):
This looks good to me, but both commits seem to change both sentences, so maybe you want to squash before merging?

src/tech/how-we-do-development.md
```diff
@@ -163,9 +163,8 @@ origin [BRANCHNAME]`.
 
 ### 8. Submit the changeset for code review
 
-At dxw, we never make changes to code running in production without the involvement of at least two people. In practice, this means one author and at least one approving reviewer, or authoring code as a pair or mob. This to make
-sure we're delivering high quality software and it's an opportunity to learn
-from each other.
+At dxw, we never make changes to code running in production without the involvement of at least two people. In practice, this means one author and at least one approving reviewer, or authoring code as a pair or mob. We do this to make
+sure we're delivering high quality software and it's an opportunity to learn from each other.
```
yndajas, line 167: Maybe `...and use it as an opportunity...`?

lookupdaily: Yes i agree, that reads better - done

PR #1068, opened 2023-02-03
Replace 10000ft with productive

snim2:


yndajas (approved):
Makes sense to me, but did you want to get someone in the people team and tech ops to confirm these processes map across 1:1?

PR #1071, opened 2023-02-03
Add more detail and examples to redirect guidance

yndajas:
This explains the different uses of the two redirect methods in more detail, and provides examples of situations in which you would each one

### DIFF
--- a/src/contributing/index.md
+++ b/src/contributing/index.md
@@ -129,13 +129,27 @@ There are two approaches to redirects in the playbook and they solve different i
 
 #### jekyll-redirect-from gem
 
-A redirect_from entry can be added to a page's front matter - this works for redirecting
-one page to another, and includes keeping the same hash fragment. This is powered by the
-jekyll-redirect-from gem and requires developer assistance. This is the preferred approach
-as it keep redirects tied coupled closely to the content.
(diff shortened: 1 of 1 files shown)


snim2 (approved):
This looks good to me, and I don't want to block it from being merged, but I wonder whether we need a slightly different set of docs for colleagues who use the CMS on the README?

yndajas:
> This looks good to me, and I don't want to block it from being merged, but I wonder whether we need a slightly different set of docs for colleagues who use the CMS on the README?

Do you mean we'd move any detailed stuff for devs more into the README, and leave the rest (CMS) in the contributing guide?

snim2:
> > This looks good to me, and I don't want to block it from being merged, but I wonder whether we need a slightly different set of docs for colleagues who use the CMS on the README?
> 
> Do you mean we'd move any detailed stuff for devs more into the README, and leave the rest (CMS) in the contributing guide?

Oh that's a really good point. I guess non-technical people will look at the site and not the README?

yndajas:
> Oh that's a really good point. I guess non-technical people will look at the site and not the README?

I'd guess so! Might be worth adding an issue to review/pick this up separately?

Issue #1074, opened 2023-02-06
Search: optimize excerpt


lookupdaily:
Original search PR: #984 
Related issues: #1025 (match-counter) #1030 (optimize search)

An excerpt for a search result is often not most relevant excerpt, which may cause a user to dismiss a result as irrelevant. Currently the excerpt is generated by finding the match index for the first matching word in the search term, but does not prioritise full matches (likely to be a more relevant result).

When you search for multiple terms it should try to find a match for the whole search phrase first, then fall back on matches for constituent terms if there's no whole-phrase match. This should make it a little clearer when results are actually relevant, but just part of very long pages.

This is linked somewhat to #1025, which suggests a match-counter for each result so that the user will be able to easily see that there are other relevant results on the same page if the except text is not what the user is looking for.

Issue #1030, opened 2022-11-25
Search: optimise results


yndajas:
Original search PR: #984
Related issue: #1024 (analytics)

Midway through development, it was noted that the relevance sorting of search results may be sub-optimal for some particular searches. For example, searching for "sickness policy", you need to scroll down to the fourth result to find the correct page

We'll need to have a think about/research whether this is a symptom of the search itself or the way things are structured. As mentioned in #1024, we could look to tweak the structure of content for internal SEO, or we could look into how to boost certain results for particular search queries

The search engine, [Elasticlunr.js](http://elasticlunr.com), has a "query-time boosting" feature, which we may be able to tap into to boost (or maybe deboost?) certain terms. For example, if "policy" is combined with another word, we might prioritise the other word since "policy" is likely more generic. I believe the engine already weights words based on how frequently they appear across the entire dataset/corpus, but boosting may be a way of augmenting this

PR #1075, opened 2023-02-07
Optimise search excerpt to highlight more relevant results

lookupdaily:
Currently a search result excerpt is generated by searching for the index of each separate term and selecting the first match for the first word (or the first match for the second if it is not found). This means that when selecting the first excerpt from the page it may not be the most relevant to the user as it may contain the first term only when there is a result lower down the page which contains the full term. (Issue #1074 )

This change looks for an excerpt containing the full search term first, so a full match will be prioritised over one containing just one word.

**A few considerations:**
- This doesn't work for special characters. However the search functionality itself does not find words with special characters in the search term (e.g. 'thoughtbot's' will not be found), so we might want to consider ignoring special characters as part of #1030 
- This is only searching for the full search term, and then each separate term in isolation. If, for example, somebody searched with 3 terms it will look for an excerpt containing 3 terms together before falling back on each term in isolation rather than combinations (e.g. the first two terms) - but not sure that doing this will add much value.

Resolves issue #1074

### DIFF
--- a/src/assets/js/search.js
+++ b/src/assets/js/search.js
@@ -98,13 +98,13 @@
   }
 
   const getExcerpt = (content, searchQuery) => {
-    const searchTerms = searchQuery.split(' ').filter(i => i).map(term => {
-      const regex = new RegExp(term, 'ig')
+    const fullSearchTerm = getTermMatcher(searchQuery, content)
+    const searchTerms = searchQuery
+      .split(' ')
+      .filter((i) => i)
+      .map((term) => getTermMatcher(term, content))
 
-      return { term, regex, matchIndex: content.search(regex) }
-    })
-
-    const firstMatchedTerm = searchTerms.find(term => term.matchIndex !== -1)
+    const firstMatchedTerm = fullSearchTerm.matchIndex !== -1 ? fullSearchTerm : searchTerms.find(term => term.matchIndex !== -1)
     const matchIndex = firstMatchedTerm ? firstMatchedTerm.matchIndex : 100
     const matchLength = firstMatchedTerm?.term.length || 0
     const excerptStartIndex = getStartIndex(matchIndex, content)
@@ -122,6 +122,11 @@
     })
 
     return excerpt
+  }
+
+  const getTermMatcher = (term, content) => {
+    const regex = new RegExp(term, 'ig')
+    return { term, regex, matchIndex: content.search(regex) }
   }
 
   const getStartIndex = (matchIndex, content) => {
--- a/src/assets/js/search.js
+++ b/src/assets/js/search.js
@@ -69,7 +69,7 @@
 
     let innerHtml = ''
 
-    results.forEach((result) => {
+    results.forEach(result => {
       const item = window.store[result.ref]
 
       const breadcrumbs = item.url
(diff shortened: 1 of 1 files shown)


src/assets/js/search.js
```diff
@@ -98,13 +98,20 @@
...
+    }
+
     const searchTerms = searchQuery.split(' ').filter(i => i).map(term => {
       const regex = new RegExp(term, 'ig')
 
       return { term, regex, matchIndex: content.search(regex) }
     })
 
-    const firstMatchedTerm = searchTerms.find(term => term.matchIndex !== -1)
+    const firstMatchedTerm = fullSearchTerm.matchIndex !== -1 ? fullSearchTerm : searchTerms.find(term => term.matchIndex !== -1)
```
yndajas, line 114: I wonder if it would be simpler to parse this if we simply added `fullSearchTerm` to the start of `searchTerms` using `.unshift`? Then the `firstMatchedTerm` line could stay simple and the hash building is a little DRYer. The result is basically the same though, so it's no big deal

```javascript
const searchTermStrings = searchQuery.split(' ').filter(i => i)

searchTermStrings.unshift(searchQuery)

const searchTerms = searchTermStrings.map(term => {
```

yndajas (approved):
Nice! This seems to be doing the job - tested with the "flexible working example", comparing the excerpt in the sixth result with that on the live version of the playbook

I've included a suggestion for DRYing this up, but I think it's good to merge in principle

## Before

image

## After

image

src/assets/js/search.js
```diff
@@ -98,18 +98,11 @@
...
-    const searchTerms = searchQuery.split(' ').filter(i => i).map(term => {
-      const regex = new RegExp(term, 'ig')
-
-      return { term, regex, matchIndex: content.search(regex) }
-    })
+    const fullSearchTerm = getTermMatcher(searchQuery, content)
+    const searchTerms = searchQuery
+      .split(" ")
+      .filter((i) => i)
+      .map((term) => getTermMatcher(term, content));
```
yndajas, line 105: Per [this recent comment](https://github.com/dxw/playbook/pull/1055#discussion_r1095849745), it would be good to keep new code in line with [Standard](https://standardjs.com) formatting guidelines, so here that would mean using single quotes and dropping the semi-colon

If you're using VSCode, there's an extension and some settings that I can share that help make this automated on save - I had to play around quite a bit to get it working properly!

yndajas, line 105: Was going to suggest that we keep things consistent in terms of not parenthesising single arguments passed to iterators, following the approach on lines 79 and 117, but I then noticed lines 31 and 72 do the opposite 🤦🏼‍♀️

Seems like Standard doesn't care about this, but maybe in a separate commit it would be good to take a stance either way (I have no strong preference!)

src/assets/js/search.js
```diff
@@ -155,6 +148,11 @@
     return endIndex
   }
 
+  const getTermMatcher = (term, content) => {
+    const regex = new RegExp(term, 'ig')
+    return { term, regex, matchIndex: content.search(regex) }
+  }
+
```
yndajas, line 155: I think this can go above `getStartIndex` - the approach to ordering methods I was going for in this file was chronological, and `getTermMatcher` is called before the index methods

yndajas (approved):
Looks great to me! This is a nice DRY up. Just a few quick comments on superficial things, but otherwise I think it's good to go

The first two commits could probably be squashed, but it does tell a story as is, so I'm easy

src/assets/js/search.js
lookupdaily: Thanks for sharing this, have reformatted using standard, and also squashed the commits

src/assets/js/search.js
lookupdaily: Done :)

src/assets/js/search.js
lookupdaily: Done. I have kept it to removing the parenthesis for single arguments, as I think it is a bit easier to read when on a single line. Line 31 does need a parenthesis though for the array.

PR #1079, opened 2023-02-10
Upgrade to Ruby 3

yndajas:
We rolled back to Ruby 2 in 5ae86e0 due to some exceptions. The specific exceptions weren't documented, but all appears to be working well after the recent Jekyll upgrade

To install Ruby 3.2.1 locally, I first had to install `libyaml` with `brew install libyaml`, per this discussion: https://github.com/rbenv/ruby-build/discussions/2118#discussioncomment-4505182


snim2:
@yndajas is it worth adding those packages to the `Brewfile` here?

yndajas:
> @yndajas is it worth adding those packages to the `Brewfile` here?

Done!

PR #1082, opened 2023-02-16
Add babel and webpack

lookupdaily:
This change is associated with issue #1031  which suggests adding lint checks for JavaScript files to the CI pipeline, but there are a couple of existing lint errors on the code which will cause a blocker for merging PRs - particularly undeclared functions from elasticlunr.

This change adds basic webpack and babel configuration to help resolve these linting errors. Modules can now be imported into JavaScript files rather than loading precompiled libraries using a script tag.

Changes include:

- set up webpack and babel to compile JavaScript files using module imports into vanilla javascript
- move JavaScript files to a new _webpack directory, which will be emitted to the same location that JavaScript files were previously saved: `src/assets/js`. This means that new JavaScript should be added to _webpack. 
- install elasticlunr import as an npm module to be imported directly into search.js, resolving linting errors
- Update serve scripts and add build script with babel
- Add a netlify.toml file to customize the build command (run webpack before jekyll build)
- Resolve linting errors on redirect.js

**Related PR:** 
It would be better to merge this change before merging the update to add linting into the CI pipeline, however have opened PR #1083 with this change implemented for reference.

**To consider:**
- These two changes introduce npm scripts for managing JavaScript files and there is now a bit of overlap/repetition between 'script/' and npm scripts.
- I had some difficulty ensuring webpack bundled js assets before `jekyll serve` for local development, and so have added a timer (sleep 5). This does not seem the best solution, but running the tasks in parallel (&) or piping (|) did not seem to always emit the js assets in time, and we cannot use sequential commands if we want to watch for JavaScript changes using `webpack --watch`.

Related to issue #1031


richpjames (approved):
I don't have much context of the Playbook but these changes make sense to me.

src/_webpack/search.js
```diff
@@ -1,3 +1,6 @@
+const elasticlunr = require('elasticlunr')
+const lunr = require('elasticlunr');
```
yndajas, line 2: Haven't looked at this properly (exciting changes though!), but if this is otherwise ready to go, would be good to remove the semicolon here to keep it Standard

Gweaton: Is it possible to configure Standard to prefer semicolons? I'm the type who likes semicolons in my JavaScript (and believe it's the default with ESLint), but maybe I need to get with the times...

Gweaton: Hah, [apparently not](https://standardjs.com/#i-disagree-with-rule-x-can-you-change-it)

lookupdaily: Love the pro tip! This semicolon is actually required because of the iife and passes the checks. An option would be to rewrite this as a module

Gweaton (approved):
This looks great, thanks for sorting!

I don't think there's any problem with `script/server` calling `npm start`. It's really just a convention across the majority of our projects for people to standardised script to run.

I had a little dig into the `sleep` issue. I'm surprised there haven't been lots of other people encountering this but did find a blog post where they do the same thing. I ran it locally and it's not really a noticeable wait, so I think this is fine for now.

src/_webpack/search.js
lookupdaily: I think I'll merge this with the semicolon for now - but happy to pick up rewriting to remove the need for a semicolon in a separate PR

yndajas: > Hah, [apparently not](https://standardjs.com/#i-disagree-with-rule-x-can-you-change-it)

Ha, yeah and I think we already chose Standard over ESLint (or semistantard) as an organisation

yndajas: > Love the pro tip! This semicolon is actually required because of the iife and passes the checks. An option would be to rewrite this as a module

Oh that's interesting. So it's different to the line before, even though they require the same thing?

I can't remember which is the more modern syntax, but is this better than `import elasticlunr as lunr from elasticlunr` or `import { elasticlunr }...`?

lookupdaily: You're right `import` is the more modern syntax. I have updated this which seems to have negated the need for the semi colon.

Elasticlunr is the default export, and I cannot import lunr for the use of `lunr.stemmer` - but there seems to be a stemmer function available on elasticlunr. Not sure on the context of why lunr was used originally here, but I have tested search and it seems to be working fine, but would be great if either of you can give search a quick test locally  to check.

Gweaton: Have tested locally too and all seems to be as expected

richpjames: > Ha, yeah and I think we already chose Standard over ESLint (or semistantard) as an organisation
From memory the reason we settled on Standard was something like 'we use it for Ruby so lets give it a go for JS'. If we're finding it's not consistently linting in a way that works for us I suggest we use something else EG semistandard

Issue #977, opened 2022-10-05
Should we run the link checker in a script


snim2:
This is a follow up to [this conversation](https://github.com/dxw/playbook/pull/973#discussion_r986669118) in #973.

Currently we advise developers to run [`html-proofer`](https://github.com/gjtorikian/html-proofer) manually, and it isn't in a `script/test` file. This is partly because we have some links to private information, which the link checker cannot follow, and we want to ignore those false-positives. However, we have missed some genuine dead links (see: #974, #976).

I've had a go at a script that would deal with false-positives, and the code below works OK at the moment (the remaining errors it currently finds are documented in #974 and #976). However, it will add a small maintenance burden as time goes on.

Note that the two private GitHub repos that are referenced could potentially be replaced with links to public repos. `wpscan.com` could be removed as a special case if we ignore HTTP 5xx status codes.

```shell
#!/bin/sh
#
# Usage:
#    script/test
#
# You can also pass extra CLI arguments to htmlproofer:
#    script/test --extra-argument

set -e

# Some URLs with fragments used for highlighting text (for example,
# https://...#:~:text=Shared%20Parental) fail the lint, but return an OK 200
# status code. These should not be reported as errors.
IGNORE_HTTP_STATUS="200"

# This trick comes from the Jekyll codebase:
#     https://github.com/jekyll/jekyll/blob/master/script/proof
IGNORE_HREFS=$(ruby -e 'puts %w{
    airtable.com
    app.productive.com
    docs.google.com
    drive.google.com
    dxw.slack.com
    github.com/dxw/judiciary-middleware
    github.com/dxw/mind-side-by-side
    wpscan.com
}.map{|h| "/#{h}/"}.join(",")')

# Build the Jekyll site here.
DESTINATION="./_site"

echo "Removing any existing version of the built site..."
rm -rf "${DESTINATION}"

echo "Building site..."
bundle exec jekyll build \
    --verbose \
    --destination ${DESTINATION} \
    --strict_front_matter \
    --trace

echo "Running htmlproofer..."
bundle exec htmlproofer ${DESTINATION} \
    --ignore-urls "$IGNORE_HREFS" \
    --ignore-status-codes "$IGNORE_HTTP_STATUS" \
    "$@"
```

PR #1090, opened 2023-02-23
Add code style checks into CI pipeline

lookupdaily:
This change adds scripts for linting JavaScript files (#1031 ) using JavaScript Standard Style, and enforces these checks on new pull requests.

- Add new npm script for linting, and include this in a new script/test script - which can include future tests and html proofing (#977 ).
- Add new GitHub workflow for runnning lint checks on creating pull requests to main branch.
- Add documentation on code style to project README.md

A PR was opened for this but the build failed after rebasing following an update to the Netlify build config. Subsequent pull requests and merges have resulted in a successful build, and I believe it may be to do with Netlify caching installed gems.

Related PR: #1032 - (which is now merged) resolved any outstanding standard lint errors, with the addition of webpack to allow module imports into JavaScript files.

Resolves issue #1031

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,19 @@
+name: Code style
+on:
+  pull_request:
+    branches:
+      - main
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    name: Lint
+    steps:
+    - uses: actions/checkout@v3
+      name: Check out repository
+    - uses: actions/setup-node@v3
+      name: Set up Node.js
+      with:
+        node-version: 18.13.0
+    - run: |
+        npm ci
+        npm run lint
--- a/README.md
+++ b/README.md
@@ -52,6 +52,21 @@ are no broken links. The following are ignored:
 - external links that require authentication, when listed in `lib/utils/check-html`
 - 200 OK status codes - this prevents false positives such as links with anchors
 
+### Code style
+
+This project uses [JavaScript Standard Style](https://standardjs.com/) to enforce consistent style. This is enforced on pull requests to the main branch. Linting is run across all JavaScript files as part of the test script.
+
+```
+script/test
+```
+
+Alternatively you can run standalone linting checks by running `npm run lint`, or fix files using `npm run lint:fix`. Specific files and directories written in the glob pattern can be passed as optional arguments to either script when surrounded by quotes.
+
+E.g.
+```
+npm run lint "src/_webpack/**/*.js"
+```
+
 ## Licence
 
 The contents of dxw's Playbook is released under a
--- a/script/test
+++ b/script/test
@@ -13,4 +13,7 @@ fi
 echo "==> Updating..."
 script/update
 
+echo "==> Linting files..."
+npm run lint
+
 script/utils/check-html
--- a/package.json
+++ b/package.json
@@ -7,7 +7,9 @@
     "webpack:dev": "webpack --mode development --watch",
     "webpack:prod": "webpack --mode production",
     "jekyll:dev": "sleep 5 && bundle exec jekyll serve",
-    "jekyll:build": "bundle exec jekyll build"
+    "jekyll:build": "bundle exec jekyll build",
+    "lint": "standard",
+    "lint:fix": "standard --fix"
   },
   "devDependencies": {
     "@babel/core": "^7.20.12",


script/test
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+
+# script/test: Run test suite for application. Optionally pass in a path to an
+#              individual test file to run a single test.
```
yndajas, line 4: Is this true? I don't think it's currently set up to handle individual files

script/test
```diff
@@ -0,0 +1,11 @@
...
+
+set -e
+
+cd "$(dirname "$0")/.."
+
+echo "Linting files..."
```
yndajas, line 10: I think we typically use the format `==> message` when `echo`ing in `script/` files. [script/bootstrap](https://github.com/dxw/playbook/blob/main/script/bootstrap) and [script/setup](https://github.com/dxw/playbook/blob/main/script/setup) have a few examples

.github/workflows/lint.yml
```diff
@@ -0,0 +1,21 @@
...
+    - uses: actions/checkout@v2
+      name: Check out repository
+    - uses: actions/setup-node@v1
+      name: Set up Node.js
+      with:
+        node-version: 18.13.0
+    - run: |
+        npm ci
+        npm run lint
+      name: Lint
```
yndajas, line 20: I don't know anything about GitHub workflows, so it might be good for someone else to look at this commit - it makes sense though 😅

README.md
```diff
@@ -53,6 +53,20 @@ bundle exec jekyll build
 bundle exec htmlproofer _site
 ```
 
+### Code style
+
+[JavaScript Standard Style](https://standardjs.com/) is the authority on style for JavaScript files. This is enforced on pull requests to the main branch. Linting is run across all files as part of the test script.
```
yndajas, line 58: `JavaScript Standard Style](https://standardjs.com/) is the authority on style for JavaScript files` - is this true? I don't know what's more popular out of Standard and ESLint. I think Standard is our standard but I'm no longer sure how standard a standard it is 😅

It's listed as an example of a linter that enforces a standard style [here](https://github.com/dxw/tech-team-rfcs/blob/main/rfc-035-use-standard-linting-tools-across-all-our-projects.md), but it being an _example_ suggests that it's not the only option - the RFC is more about using some kind of standard rather than the Standard standard specifically

yndajas (approved):
Looks good. Just a few small details, and a suggestion for someone else to review the workflow part

README.md
lookupdaily: Sorry I've used funny wording there - I meant on this project. I'll reword that so it sounds less like **the** authority!

script/test
lookupdaily: Thank you for spotting this, thats a line i copied from the script/ templates and didn't come back to. I can set this up fairly easily. But wondering if setting it up will become an issue if we add other test suites to this script? 

e.g. @snim2 I know you have been looking at adding html-proofer to a test suite, which I think doesn't accept file paths as far as I can see? Will it be confusing if arguments can be passed in which are only relevant for different sections of the test script? 

I wonder if its simpler to keep a standard script for the whole suite, and developers have the tools to run certain commands individually using this as a reference.

snim2: It's totally fine to have arguments that are used in some places but not others. We usually add both testing and linting to these scripts, so it's quite common.

You may already have [read this](https://github.com/github/scripts-to-rule-them-all), but if not it's a good introduction to this design pattern.

lookupdaily: Thanks Sarah. I have updated this now, and it can also accept standard flags, such as --fix, which is explained in the if/else.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,21 @@
...
+    name: Lint
+    steps:
+    - uses: actions/checkout@v2
+      name: Check out repository
+    - uses: actions/setup-node@v1
+      name: Set up Node.js
+      with:
+        node-version: 18.13.0
+    - run: |
+        npm ci
```
snim2, line 18: Good to see `npm ci` here and not `npm install` :)

.github/workflows/lint.yml
```diff
@@ -0,0 +1,21 @@
...
+on:
+  pull_request:
+    branches:
+      - main
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    name: Lint
+    steps:
+    - uses: actions/checkout@v2
```
snim2, line 11: I think the latest version of this is >3

.github/workflows/lint.yml
snim2: Do we need that last line?

README.md
```diff
@@ -53,6 +53,21 @@ bundle exec jekyll build
...
+```
+script/test
+```
+
+Alternatively you can run standalone linting checks by running `npm run lint`, or fix files using `npm run lint:fix`. Specific files and directories written in the glob pattern can be passed as optional arguments to either script when surrounded by quotes.
+
+E.g.
+```
+npm run lint "src/_webpack/**/*.js"
+```
```
snim2, line 69:
```suggestion
E.g.

```
npm run lint "src/_webpack/**/*.js"
```
```

script/test
snim2: OK, but does the comment at the top of the file match the behaviour?

snim2:
Thanks @lookupdaily! This looks good. There's a deprecation warning that you might want to deal with while you are here:

```
Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: actions/checkout@v2, actions/setup-node@v1. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.
```

but maybe it will fix itself if `actions/checkout` is updated, or when `ubuntu-latest` rolls over in April?

.github/workflows/lint.yml
lookupdaily: Updated both steps v3

script/test
lookupdaily: Yes, you can now pass in one file or a directory - which works for this implementation but we may need to keep an eye on if adding test suites or proofing to the script. Have kept specifics of standards arguments to the section where linting is run, but let me know if you think it would be better to move that to the top

script/test
```diff
@@ -14,13 +14,6 @@ echo "==> Updating..."
...
 echo "==> Linting files..."
-if [ -n "$1" ]; then
-  # pass arguments to standard. This is useful for checking a single JavaScript file, 
-  # or a directory using the glob pattern. You can also pass in --fix to fix 
-  # lint errors.
-  npx standard "$@"
-else
-  npx standard
-fi
+npm run lint
```
yndajas: How does `npm run lint` know not to care about the bundled scripts in `src/assets/js`? It doesn't complain about them, just curious why not 🤔 (maybe they're compliant with Standard)

yndajas (approved):
Looks good, just a couple of things:
- is it worth squashing some of the commits to keep a clean git history? A few of the commits undo things from previous commits, so it might be good to simplify it before merging into main, for ease of reference in future
- the query below

script/test
lookupdaily: Good question! Just found this in the docs, which is nice! [https://standardjs.com/#how-do-i-ignore-files](https://standardjs.com/#how-do-i-ignore-files)

> Certain paths (node_modules/, coverage/, vendor/, *.min.js, and files/folders that begin with . like .git/) are automatically ignored.
> 
> **Paths in a project's root .gitignore file are also automatically ignored.**

yndajas: Nice!

Issue #1092, opened 2023-02-27
Find a way to include responsive accessible images


snim2:
PR #1091 should have had some images included in it, but we haven't chosen a set of plugins for creating responsive, accessible images. The ones I've seen seem to use Imagemagick on the command-line to resize images, and that won't work nicely in the CMS, so we'd need to build something more automated into the pipeline.

That's more time than I have available right now, and it needs some design discussion.

yndajas:
These PRs from the accessibility handbook may be relevant:
- https://github.com/dxw/accessibility-manual/pull/21
- https://github.com/dxw/accessibility-manual/pull/26
- https://github.com/dxw/accessibility-manual/pull/28

I believe we now have image upload working (although we haven't properly tested on production since merging PR 26) and have basic image width constraints in place*, which should make them vaguely responsive. You can add alt text in the UI

\* actually PR 28 doesn't seem to have deployed correctly 🤔

snim2:
@yndajas thanks for this! I'll have a go at raising a PR and if it all looks good I'll close this one 🎉

yndajas:
You should see it as a field in the image upload widget:

<img width="692" alt="image" src="https://user-images.githubusercontent.com/40244233/221652884-9cd39a7e-c7ea-4465-8301-281d30113a26.png">

snim2:
Ah, sorry! `![THIS TEXT](...)` did produce ALT text, I just missed it outside of the Inspector. Will close.

PR #1095, opened 2023-02-27
Enable image uploads

yndajas:
This follows work on the accessibility handbook as detailed in [this comment](https://github.com/dxw/playbook/issues/1092#issuecomment-1446090962), enabling image uploads via the Netlify CMS, constraining them to avoid overflow (at any screen size) and shifting existing images into relevant folders

Addresses #1092 (hopefully)


snim2 (approved):
Thanks for doing this!

PR #1144, opened 2023-04-14
Create new page on our use of data standards

thill-dxw:
Added a new page outlining the appropriate use of data standards, focusing on:

1. Standards mandated by the Service Standard
2. Metadata standards recommended for use by GOV.UK
3. GSS Harmonised Standards
4. DSA-endorsed standards

### DIFF
--- a/src/tech/data-standards.md
+++ b/src/tech/data-standards.md
@@ -1,0 +1,45 @@
+# Data Standards at dxw
+
+In keeping with the GOV.UK [Service Manual](https://www.gov.uk/service-manual/service-standard/point-13-use-common-standards-components-patterns), dxw adheres to relevant open data standards wherever possible. While such standards are often recommended or mandated by government requirements, they are frequently useful in private and third-sector work as well.
+
+## Why standards are important
+
+There are three chief reasons why using data standards makes sense.
+
+First, organisations - and in particular, government - are becoming ever more aware of the importance of data. The advent of machine learning and artificial intelligence technologies, alongside the steadily growing importance of statistics in evidence-driven decision- and policy-making, means that not only survey data, but also routine administrative data, need to be treated as a potentially valuable resource. Care needs to be taken to ensure that data is collected in ways that ensure it is consistent, coherent, and comparable across the whole of UK public services.
+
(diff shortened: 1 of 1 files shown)


src/tech/data-standards.md
```diff
@@ -0,0 +1,45 @@
+# Data Standards at dxw
```
yndajas, line 1: Some of the Jekyll front matter is missing here. At a minimum:

```
---
title: Data standard at dxw
permalink: /:path/:basename/
---

```

Then the `h1` can go away (I sentence cased it in the example above because most of our playbook pages have sentence case titles)

thill-dxw: Good catch! Fixed.

src/tech/data-standards.md
```diff
@@ -0,0 +1,47 @@
...
+
+Finally, data standards can help systems interoperate, reducing the risk of silos and the need for middleware to bridge gaps in information exchange.
+
+## Which standard to use
+
+Choosing and/or finding a relevant data standard can be difficult. As the old joke goes, _standards are great – that's why there are so many of them_. This is particularly the case in government, where particular ministries or councils will often develop and promote their own bespoke ways to capture data _ad hoc_ or to suit their own very particular needs.
+
+### Mandated standards
+
+For some well-defined areas, the UK government mandates the use of [particular standards](https://www.gov.uk/government/publications/open-standards-for-government). These are all widely used and well-tested and are suitable for use on almost any project, whether for governmental or other organisations.
```
jdudley1123, line 23: 'government' in 'UK government' should be capitalised to be consistent with the next section

src/tech/data-standards.md
```diff
@@ -0,0 +1,47 @@
...
+
+## When not to use standards
+
+All standards mandated or recommended by the UK Government have been extensively tested and their value proven 'in the wild'. Under most circumstances, then, their use provides an easy and effective means of improving data collection for our clients.
+
+Some services and applications, however, will have specialised or unique requirements unsuitable for standardisation. For instance, standard age-bandings may not be useful to a service focused on Early Years Foundation Stage education; here, finer-grained measures will probably be needed. Be sure to consult with relevant stakeholders before implementing a standard to confirm that it does in fact meet their needs. Even in cases where a standard as defined cannot be directly applied, it will often be possible to create data structures that can be mapped to it without information loss; for instance, in the age-banding example given above, care could be taken to ensure that the more-specific groupings defined fall within the boundaries of one of the less-granular standards.
```
jdudley1123, line 45: I would make this point a little more strongly. ONS/GSS/GDS standards tend to be designed for maximum clarity for the general public, whereas many of the things we build the users are internal specialists or power users for whom different standards or nomenclature might be more appropriate – for example the police use [IC codes](https://en.wikipedia.org/wiki/IC_codes) to describe ethnicities.

jdudley1123 (approved):
Looks good!

src/tech/data-standards.md
thill-dxw: I've expanded slightly to emphasise that the data standards described are 'sensible defaults' rather than 'universal solutions'. I haven't referred to the IC codes specifically, partly because the Wikipedia entry goes into some detail about the relationship of IC codes to other standards and the example started seeming less clear-cut than desirable.

PR #1101, opened 2023-02-28
Content restructure

yndajas:
The bulk of this PR is a big content restructure done by Dawn, introducing a new staff handbook section. It furthers the work F did in #973, breaking down content into more logical sections, making the playbook easier to navigate and consume. This has been approved by both Pauline (on behalf of HR) and Wendy (on behalf of marketing and communications), so it mostly needs a technical review/check before merging

PR #1145, opened 2023-04-18
Fix Netlify CMS config for staff handbook

yndajas:
This was missed in #1101, where we moved from a "Working here" section to a "Staff handbook" section as part of a content restructure. This config is required in order to make the section's pages editable via the Netlify CMS


snim2 (approved):
lgtm thanks

PR #1148, opened 2023-05-04
Who we are content restructure

yndajas:
Dawn's work on restructuring the "Who we are" section, including renaming it to "About us"


src/about-us/how-dxw-is-run.md
```diff
@@ -0,0 +1,4 @@
+---
```
snim2: There are three empty files added in the first commit, was that intended?

yndajas: I think Dawn was creating the structure in one commit and the content in another. We could rebase this, but given content changes often come in a series of non-atomic commits from the CMS, I'm tempted not to bother in this context. Happy to squash some of the commits though if preferable

src/about-us/who-we-are.md
```diff
@@ -11,4 +11,22 @@ Our approach is to add value early and often, rapidly researching, testing, lear
...
+
+Sales: generates a pipeline of work that’s on mission and provides work for the people we have.
+
+Delivery+: specialises in creating bespoke digital transactional services by providing multidisciplinary teams
+
+GovPress: designs, builds, hosts and supports content-first websites for public and third sector clients
+
+Strategy: helps organisations form strategies, plans and business cases, and build the capability needed for digital transformation
+
+Product and Performance
```
snim2: Have we missed some text here?

yndajas: (I am tempted to do some squashing here but I really don't have the dxw time!)

yndajas:
> This is not easy to review, but I think it's good to merge!

As discussed, have squashed Dawn's eight content commits and my four tidy up commits into a single summarised commit to keep the history more straightforward for future reference. Thanks for the review!

Issue #1086, opened 2023-02-21
Redirects for user research guidance


johnwaterworth:
We've moved and restructured the user research guidance the Playbook.

To make sure that previous links keep working, we need at least:

redirect from:
  /guides/how-we-do-user-research#how-we-protect-participant-privacy
  /user-research/how-we-do-user-research#how-we-protect-participant-privacy
to:
  /user-research/protecting-participant-privacy

redirect from:
  /guides/how-we-do-user-research#how-we-write-and-use-discussion-guides
  /user-research/how-we-do-user-research#how-we-write-and-use-discussion-guides
to:
  /user-research/writing-and-using-discussion-guides

redirect from:
  /guides/how-we-do-user-research#how-we-get-informed-consent-for-user-research
  /user-research/how-we-do-user-research#how-we-get-informed-consent-for-user-research
to:
  /user-research/getting-informed-consent-for-research

redirect from:
  /guides/how-we-do-user-research#how-we-create-and-use-research-plans
  /user-research/how-we-do-user-research#how-we-create-and-use-research-plans
to:
  /user-research/creating-and-using-research-plans

redirect from:
  /guides/how-we-do-user-research#how-we-take-notes-and-record-research-sessions
  /user-research/how-we-do-user-research#how-we-take-notes-and-record-research-sessions
to:
  /user-research/taking-notes-and-recording-research-sessions

redirect from:
  /guides/how-we-do-user-research#how-we-share-research-activities-and-findings
  /user-research/how-we-do-user-research#how-we-share-research-activities-and-findings
to:
  /user-research/sharing-research-activities-and-findings

redirect from:
  /guides/how-we-do-user-research#how-we-name-and-identify-documents
  /user-research/how-we-do-user-research#how-we-name-and-identify-documents
to:
  /user-research/naming-and-identifying-research-documents

redirect from:
  /guides/how-we-do-user-research#when-and-how-we-give-incentives-to-research-participants
  /user-research/how-we-do-user-research#when-and-how-we-give-incentives-to-research-participants
to:
  /user-research/when-and-how-to-give-incentives-to-research-participants

redirect from:
  /guides/how-we-do-user-research#how-we-do-research-safely-during-coronavirus
  /user-research/how-we-do-user-research#how-we-do-research-safely-during-coronavirus
to:
  /user-research/doing-research-safely

redirect from:
  /guides/how-we-do-user-research#how-we-take-care-of-research-data
[...]

PR #1149, opened 2023-05-10
Check and fix links

### LINKED ISSUE #977: Should we run the link checker in a script

yndajas:
This fixes a bunch of internal 404s, mostly from the working here -> staff handbook restructure. It also updates some `/working-here` links that weren't 404s thanks to redirects being set up. There's also one casing change for consistency, where this section (mental health first aiders) was being linked to

Also fixed are two external 404s - the rest were false positives

Lastly, a script is added (mostly taken from #977) to run the HTML checker, ignoring known false positives. This is not yet integrated into the CI - doing so may create extra overhead if non-technical staff make edits that introduce false positives (see discussion below, in #977 and the Slack conversation linked therein)

Closes #977

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,12 +46,11 @@ This runs the server on localhost:4000.
 
 ### Checking the HTML
 
-To proof the HTML for things like dead links, run:
+After making a change, please run `script/test`, which will ensure that there
+are no broken links. The following are ignored:
 
-```
-bundle exec jekyll build
-bundle exec htmlproofer _site
-```
+- external links that require authentication, when listed in `lib/utils/check-html`
+- 200 OK status codes - this prevents false positives such as links with anchors
 
 ## Licence
 
--- a/script/test
+++ b/script/test
@@ -4,4 +4,13 @@
 
 set -e
 
+cd "$(dirname "$0")/.."
+
+if [ -n "$DEBUG" ]; then
+  set -x
+fi
+
+echo "==> Updating..."
+script/update
+
 script/utils/check-html
--- a/script/test
+++ b/script/test
@@ -1,0 +1,7 @@
+#!/bin/sh
+
+# script/test: Run the test suite for the application.
+
+set -e
+
+script/utils/check-html
--- a/script/update
+++ b/script/update
@@ -1,0 +1,10 @@
+#!/bin/sh
+
+# script/update: Update the application to run for its current checkout.
+
+set -e
+
+cd "$(dirname "$0")/.."
+
+echo "==> Bootstrapping..."
+script/bootstrap
--- a/script/utils/check-html
+++ b/script/utils/check-html
@@ -1,0 +1,49 @@
+#!/bin/sh
+
+# script/utils/check-html: Validate the HTML built by Jekyll for issues with
+#                          images, links and scripts (including broken links).
+
+#                          This will delete any existing build in `_site`
+#                          and build the site fresh.
+
+#                          This uses htmlproofer and you can pass additional
+#                          htmlproofer arguments like so:
+#                            script/utils/check-html --extra-argument
+
+set -e
+
+# Some URLs containing fragments for highlighting text (for example,
+# https://...#:~:text=Shared%20Parental) or navigating to anchors (including
+# specific slides e.g. https://...#slide=id.g5fe190064c_6_403fail) fail the
+# lint, but return a 200 OK status code. These should not be reported as errors.
+IGNORE_HTTP_STATUS="200"
+
+# Some URLs are authenticated or fail for some other reason that isn't an issue.
+# The syntax comes from the Jekyll codebase:
+#   https://github.com/jekyll/jekyll/blob/master/script/proof
+IGNORE_HREFS=$(ruby -e 'puts %w{
+  expenses.xero.com/!wrUP-/detail/create-new/
+  github.com/dxw/govpress-developer-docs
+  github.com/dxw/judiciary-middleware
+  github.com/dxw/mind-side-by-side
+  mheducation.co.uk/the-good-research-guide-research-methods-for-small-scale-social-research-projects-9780335249831-emea-group
+  twitter.com
+}.map{ |url| "/#{url}/" }.join(",")')
+
+BUILD_PATH="./_site"
+JS_ASSET_PATH="./src/assets/js"
+
+echo "==> Removing any existing build..."
+
+rm -rf "${BUILD_PATH}" "${JS_ASSET_PATH}"
+
+echo "==> Building site..."
+
+npm run build
+
+echo "==> Checking the HTML"
+
+bundle exec htmlproofer ${BUILD_PATH} \
+    --ignore-status-codes "$IGNORE_HTTP_STATUS" \
+    --ignore-urls "$IGNORE_HREFS" \
+    "$@"
--- a/src/contributing/managing-the-playbook.md
+++ b/src/contributing/managing-the-playbook.md
@@ -90,7 +90,7 @@ The owners of each section are:
   [The work we do > Supporting services](/work-we-do/supporting-services/)
 * Wendy Coello -
   [The work we do > Sharing our expertise](/work-we-do/sharing-expertise/)
-* Poss Apostolou - [Working here](/working-here/)
+* Poss Apostolou - [Staff handbook](/staff-handbook/)
 * Coca Rivas - Our professions
 
 ### Workflow
--- a/src/staff-handbook/dxw-time.md
+++ b/src/staff-handbook/dxw-time.md
@@ -41,4 +41,4 @@ If you’re in one of our billable teams, please talk to your line manager and a
 
 ## Non-billable staff
 
-Of course, people in our non-billable teams also get involved in company-wide work and [personal learning and development activity](staff-handbook/learning-and-development/) outside of their day jobs. This should be agreed with their line manager as part of their objectives in a similar way.
+Of course, people in our non-billable teams also get involved in company-wide work and [personal learning and development activity](/staff-handbook/learning-and-development/) outside of their day jobs. This should be agreed with their line manager as part of their objectives in a similar way.
--- a/src/staff-handbook/pay-pension-and-benefits/benefits.md
+++ b/src/staff-handbook/pay-pension-and-benefits/benefits.md
@@ -12,9 +12,9 @@ Benefits for permanent employees include:
 
 * [Electric vehicle scheme](https://www.electriccarscheme.com)
 
-* up to 18 weeks of paid [parental leave](/staff-handbook/policies/parental-leave-policy/) depending on your situation
+* up to 18 weeks of paid [parental leave](/staff-handbook/policies-and-procedures/parental-leave-policy/) depending on your situation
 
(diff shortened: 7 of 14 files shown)


davidwinter:
@yndajas a quick thought on the authenticated pages false positives, we could update `htmlproofer` to ignore unauthenticated http status codes (with the `--ignore-status-codes` flag) to reduce this number down?

https://github.com/gjtorikian/html-proofer#special-cases-for-the-command-line

Not necessarily for this PR, but something to consider.

yndajas:
> @yndajas a quick thought on the authenticated pages false positives, we could update `htmlproofer` to ignore unauthenticated http status codes (with the `--ignore-status-codes` flag) to reduce this number down?

I've added a couple of extra commits fixing the actually broken external links, and a script for automating this and keeping track of known false positives

I'm not 100% sure about this approach:
- we could be creating future false negatives
- does it make sense to split this out from `script/test` into a Ruby script in `lib`?

Any thoughts?

Looping in @snim2 since this now touches #977, and I've _just_ realised there's a premade more nuanced script in there already! Maybe we can fold that in here

snim2:
@yndajas so, the script in #977 always worked fine for me (it also exists in this branch: https://github.com/dxw/playbook/tree/chore/add-linting). 

Ideally, we'd run something like this in CI, but then if someone edits a page from the UI interface, then they won't easily be able to add an exception into that script if it's needed.

yndajas:
> if someone edits a page from the UI interface, then they won't easily be able to add an exception into that script if it's needed

Yup, that's another issue. I suppose a dev could drop in to push a one-line fix for a false positive, or we could set something up like the [redirect manager](https://github.com/dxw/playbook/blob/main/src/contributing/index.md#redirect-manager) for HTML proofing exceptions, but that's a lot more work

Do you think it would be useful to merge in something like your script, even if unconnected to the CI (which we can think about separately)?

I'd be tempted to break out the HTML checking part into something like `script/utils/check-html` and then reference it in `script/test`, to make it easier to expand `script/test` with other stuff, but otherwise it looks pretty ready to go? 

Materially it's much the same as https://github.com/dxw/playbook/pull/1149/commits/810c6fc86773270296a671af164b999ea00d8566, just with some extra nuance/safety and ignoring the 200 status code instead of external hashes. The latter might be necessary to ignore some Google Slides hash false positives though

snim2:
@yndajas sure, the trickery with the HTTP codes was the useful thing in that commit :) Feel free to steal it and I'll delete the branch when you're done (and in that case, you should probably mark this PR as resolving #977)

README.md
```diff
@@ -46,12 +46,18 @@ This runs the server on localhost:4000.
 
 ### Checking the HTML
 
-To proof the HTML for things like dead links, run:
-
-```
-bundle exec jekyll build
-bundle exec htmlproofer _site
-```
+To proof the HTML for things like dead links, run `script/test` or
```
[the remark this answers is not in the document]
yndajas: Looks good to me! I'll make this change when I tidy this up

yndajas:
@davidwinter @snim2 updated with an amended version of Sarah's shell script, and the README text condensed

script/test
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+# script/test: Run the test suite for the application.
+
+set -e
+
+script/utils/check-html
```
snim2, line 7: In the `scripts-to-rule-them-all` repo we usually run `script/update` at the top of the script. For this one, I get:

```shell
❯ ./script/test
==> Removing any existing build...
==> Building site...
Could not find nokogiri-1.14.3-arm64-darwin, html-proofer-5.0.6, async-2.4.2, io-event-1.1.7 in locally installed gems
Run `bundle install` to install missing gems.
```

So, I need to run `./script/setup` manually before continuing.

snim2:
@yndajas this is looking nice! This is the output I get from `./script/test`:

```shell
==> Checking the HTML
Running 3 checks (Images, Links, Scripts) in ["./_site"] on *.html files ...


Checking 379 external links
Checking 735 internal links
Checking internal link hashes in 84 files
Ran on 157 files!


For the Scripts check, the following failures were found:

* At ./_site/404.html:82:

  internal script reference /assets/js/redirect.js does not exist

* At ./_site/index.html:21:

  internal script reference /assets/js/redirect.js does not exist

* At ./_site/search.html:1154:

  internal script reference /assets/js/search.js does not exist

* At ./_site/staff-handbook/leave/parental-leave/index.html:82:

  internal script reference /assets/js/redirect.js does not exist

* At ./_site/staff-handbook/leave/sick-leave/index.html:82:

  internal script reference /assets/js/redirect.js does not exist

* At ./_site/staff-handbook/pay-pension-and-benefits/pay/index.html:82:

  internal script reference /assets/js/redirect.js does not exist


HTML-Proofer found 6 failures!
```

script/test
yndajas: Interesting! We don't do that in either of the project repos I've worked on, but I'm happy to add it. In `scripts-to-rule-them-all`, it looks like `update` is just wrapping around `bootstrap`, but I guess it might be good to include and use `update` by convention?

yndajas:
> @yndajas this is looking nice! This is the output I get from `./script/test`:

Huh, that's weird. I get:

```
==> Removing any existing build...
==> Building site...
Configuration file: /Users/yndajas/code/github.com/dxw/playbook/_config.yml
            Source: /Users/yndajas/code/github.com/dxw/playbook/src
       Destination: /Users/yndajas/code/github.com/dxw/playbook/_site
 Incremental build: disabled. Enable with --incremental
      Generating...
           Warning: Empty `slug` generated for '/'.
                    done in 0.435 seconds.
 Auto-regeneration: disabled. Use --watch to enable.
==> Checking the HTML
Running 3 checks (Images, Links, Scripts) in ["./_site"] on *.html files ...


Checking 379 external links
Checking 735 internal links
Checking internal link hashes in 84 files
Ran on 157 files!


HTML-Proofer finished successfully.
Finished in 4.76 seconds
```

snim2:
@yndajas Nice! I get this now:

```shell
❯ ./script/test
==> Removing any existing build...
==> Building site...

> build
> npm run webpack:prod && npm run jekyll:build


> webpack:prod
> webpack --mode production

asset search.js 19.6 KiB [emitted] [minimized] (name: search) 1 related asset
asset redirect.js 190 bytes [emitted] [minimized] (name: redirect)
runtime modules 663 bytes 3 modules
cacheable modules 71.8 KiB
  ./src/_webpack/search.js 4.96 KiB [built] [code generated]
  ./src/_webpack/redirect.js 329 bytes [built] [code generated]
  ./node_modules/elasticlunr/elasticlunr.js 66.6 KiB [built] [code generated]
webpack 5.76.2 compiled successfully in 476 ms

> jekyll:build
> bundle exec jekyll build

Configuration file: /Users/snim2/code/github.com/dxw/playbook/_config.yml
            Source: /Users/snim2/code/github.com/dxw/playbook/src
       Destination: /Users/snim2/code/github.com/dxw/playbook/_site
 Incremental build: disabled. Enable with --incremental
      Generating...
           Warning: Empty `slug` generated for '/'.
                    done in 0.737 seconds.
 Auto-regeneration: disabled. Use --watch to enable.
==> Checking the HTML
Running 3 checks (Images, Links, Scripts) in ["./_site"] on *.html files ...


Checking 379 external links
Checking 735 internal links
Checking internal link hashes in 84 files
Ran on 157 files!


HTML-Proofer finished successfully.
Finished in 4.2 seconds
```

snim2 (approved):

script/test
yndajas: Added this in a final commit

davidwinter (approved):

davidwinter:
I just found another link:

https://playbook.dxw.com/guides/guide-to-wellbeing-at-dxw

I wonder where these old URLs came from, or at what point they broke? I have encountered a fair few. Was there a major restructuring of the content at some point? Perhaps we should checkout an old version before the restructuring, grab all of the URLs, and then test those against the current playbook to catch these?

snim2:
> I just found another link:
> 
> https://playbook.dxw.com/guides/guide-to-wellbeing-at-dxw
> 
> I wonder where these old URLs came from, or at what point they broke? I have encountered a fair few. Was there a major restructuring of the content at some point? Perhaps we should checkout an old version before the restructuring, grab all of the URLs, and then test those against the current playbook to catch these?

There have been a couple of major restructurings, and I think all of the "guides" have gone now.

yndajas:
> There have been a couple of major restructurings, and I think all of the "guides" have gone now.

Yup no guides since #973. There was another major restructure in #1101. Where was that link @davidwinter? We can [add redirects in a couple of ways](https://github.com/dxw/playbook/blob/main/src/contributing/index.md#redirects) if any are found. I suppose we could go to the commit before the first in #973 to get a bunch of the old URLs that might have links hanging around outside the playbook

#1086 also lists a few that need setting up in the user research section

davidwinter:
I think from a combination of my Trello onboarding board, Google docs, historical Slack messages etc. So I think retrieving a list of old URLs might be the most helpful?

PR #1166, opened 2023-06-16
Configure VSCode to use Standard

yndajas:
Builds on #1031, which should be merged first

This:
- makes Prettier ignore everything, so that if you have the Prettier VSCode extension installed you don't need to disable it to work on JavaScript files here
- recommends the VSCode Standard extension and sets it as the default formatter


.vscode/settings.json
```diff
@@ -0,0 +1,6 @@
+{
+  "[js]": {
+    "editor.defaultFormatter": "standard.vscode-standard",
+    "editor.formatOnSave": true
```
lookupdaily, line 4: I wonder if `formatOnSave` should still be a user level setting? I have no strong opinions on it though, just took me a long time to switch to 'on save' actions!

lookupdaily (approved):
Looks good, and works for me!

.vscode/settings.json
yndajas: Probably actually, given how many different developers we have! Without it specified here, it will obviously still respect your own `formatOnSave` being on

PR #1170, opened 2023-06-29
Update plugin-reviews.md

leedxw:
GovPress hosting-specific checks

### DIFF
--- a/src/govpress-unit/plugin-reviews.md
+++ b/src/govpress-unit/plugin-reviews.md
@@ -67,25 +67,25 @@ is unsafe.
 A plugin inspection mostly involves checking these pattern matches to see if
 they represent real issues.
 
-## Checking for hosting-environment compatibility
+## Checking for hosting environment compatibility
 
 In addition to general security issues, the plugin should also be checked for
 likely incompatibilities with the hosting environment.
 
 These include, but are not limited to:
 
-* Direct calls to php files (i.e. in URLs) under `wp-content`.
-* Use of `/xmlrpc.php`.
-* Assumption the directories other than `wp-content/uploads` or
-  `wp-content/cache` can be written to (unless otherwise configurable).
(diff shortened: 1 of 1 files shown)


jkeasley (approved):
LGTM

src/govpress-unit/plugin-reviews.md
```diff
@@ -67,6 +67,26 @@ is unsafe.
...
+* Assumption the directories other than `wp-content/uploads` or
+  `wp-content/cache` can be written to (unless otherwise configurable).
+* Use of PHP sessions (check for `session_start()`)
+* Assumption that `wp_` is the database prefix.
+* Assumption that `/wp-admin/admin-ajax.php` can be accessed by non-logged-in
+  users.
+* Reliance on PHP modules other than those suggested by WordPress core.
+* Reliance on `.htaccess` files.
+* Features likely to take more than 30 seconds to output responses, even
+  in the administrative interface.
```
yndajas, line 88: Just a few minor suggestions to keep it playbook style

```suggestion
## Checking for hosting environment compatibility

In addition to general security issues, the plugin should also be checked for
likely incompatibilities with the hosting environment.

These include, but are not limited to:

* direct calls to PHP files (i.e. in URLs) under `wp-content`
* use of `/xmlrpc.php`
* an assumption that it's possible to write to directories other than
  `wp-content/uploads` or `wp-content/cache` (unless otherwise configurable)
* use of PHP sessions (check for `session_start()`)
* an assumption that `wp_` is the database prefix
* an assumption that `/wp-admin/admin-ajax.php` can be accessed by users who
  are logged out
* reliance on PHP modules other than those suggested by WordPress core
* reliance on `.htaccess` files
* features likely to take more than 30 seconds to output responses, even
  in the administrative interface
```

yndajas (approved):

PR #1174, opened 2023-07-05
Adding additional Landmark spaces to Our offices page

Chanel-Diep:
Updating the office page to include other Landmark offices we can use. 

Also to add missing 'dxw-' to Slack channel name for Leeds office.


yndajas:
These changes have already been merged - not sure what happened here

Issue #1024, opened 2022-11-25
Search: analytics


yndajas:
Original search PR: #984

This was a suggestion from the content team (Dawn and Louise, specifically). In order to inform future content changes (e.g. restructures), it might be useful to have some kind of analytics to track common search terms. My understanding is that this is taken as a proxy for "what's difficult to find via standard navigation"

Design input might be helpful here to narrow down what we want to learn and why, as well as user research (when the search feature has been around long enough) on how people find content in the Playbook (search vs menus) and why. It may be that people generally prefer search over menus, because it's quicker than working out where to look, regardless of how intuitive the menu structure is

Beyond "what's difficult to find via standard navigation", analytics could help expose potential optimisations for the search. E.g. if people regularly search "sickness policy" and click the link in the fourth result (which is currently where the page containing the policy lands), that may suggest that we should look into boosting particular results for particular search keywords/phrases (or restructure the content to optimise for internal SEO). "Query-time boosting" is a feature of ElasticLunr, the chosen search engine, but I'm not too familiar with the extent and limitations of what we can do with it

lookupdaily:
I am going to leave this open while I explore 'Conversions' in Plausible - to see if we can follow which linked somebody clicked on first (as suggested in the second point)

PR #1181, opened 2023-07-20
Capture search terms in playbook analytics

lookupdaily:
This adds the ability to track common search terms on the playbook using our analytics tool [Plausible](https://plausible.io/docs).  

This is related to issue #1024 and was a suggestion from the content team. In order to inform future content changes (e.g. restructures), it might be useful to have some kind of analytics to track common search terms. This has two benefits:
- understanding what might be difficult to find by standard navigation
- highlighting potential content gaps in the playbook

By default Plausible strips query params from urls to count page views. We can however, use a custom script to [capture query parameters](https://plausible.io/docs/custom-query-params), and track them as page views in Plausible. 

## How it works
By creating a 'Goal' for the `/search*` url in the Plausible dashboard, we can still view the total number of visitors to the search page. When clicking on the goal, common terms are listed as pages under the 'Top pages' section. Common search terms will also be listed under general page views, but this is an easy way to view them together. 

Goals shown on the Plausible Dashboard showing 'Search' goal
Top search terms shown as page views

### Special characters
This solution currently only handles spaces in search terms, and joins them together using a '-'. It does not remove other special characters. However, special characters are shown on the plausible dashboard (see example below). It is also worth noting that search itself does not currently handle special characters. This does mean that there may be more variations of the same or similar terms though - which might create additional work when analysing the top terms.
Special characters are shown in the plausible dashboard

## What's next
We are also looking ways to understand if people are finding what they're looking for on the playbook. One suggestion is to add a button on the search page to let users let us know themselves (linked to a simple form).

### DIFF
--- a/src/_layouts/default.html
+++ b/src/_layouts/default.html
@@ -23,31 +23,6 @@
 
     <link rel="stylesheet" href="/assets/main.css"/>
     <script defer data-domain="playbook.dxw.com" src="https://plausible.io/js/script.manual.js"></script>
-    <!-- define the `plausible` function to manually trigger events - as the pageview event is not being sent automatically with 'script.manual.js' -->
-    <script>
-      window.plausible = window.plausible || function () {
-        (window.plausible.q = window.plausible.q || []).push(arguments)
-      }
-    </script>
-
-    <!-- trigger pageview -->
-    <script>
-      function prepareUrl(params) {
-        const url = new URL(location.href)
-        const queryParams = new URLSearchParams(location.search)
-        let customUrl = url.protocol + "//" + url.hostname + url
-          .pathname
-          .replace(/\/$/, '')
-        for (const paramName of params) {
-          const paramValue = queryParams.get(paramName)
-          if (paramValue) {
-            customUrl = customUrl + '/' + paramValue
-          }
-        }
-        return customUrl
-      }
-      plausible('pageview', {u: prepareUrl([])})
-    </script>
   </head>
   <body>
     {% include navbar.html %}
@@ -56,5 +31,6 @@
       {% include page-content.html %}
     </main>
     {% include contribute.html %}
+    <script src="/assets/js/plausible.js"></script>
   </body>
 </html>
--- a/src/_layouts/default.html
+++ b/src/_layouts/default.html
@@ -1,11 +1,11 @@
 <!DOCTYPE html>
 <html lang="en-GB">
   <head>
-    <meta charset="utf-8" />
+    <meta charset="utf-8"/>
     <meta
       name="viewport"
       content="width=device-width, initial-scale=1, shrink-to-fit=no"
-    />
+/>
     <link rel="apple-touch-icon" sizes="180x180" href="/assets/images/favicon/apple-touch-icon.png">
     <link rel="icon" type="image/png" sizes="32x32" href="/assets/images/favicon/favicon-32x32.png">
     <link rel="icon" type="image/png" sizes="16x16" href="/assets/images/favicon/favicon-16x16.png">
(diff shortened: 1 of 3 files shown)


adamhughes86 (approved):

src/_layouts/default.html
```diff
@@ -1,11 +1,11 @@
 <!DOCTYPE html>
 <html lang="en-GB">
   <head>
-    <meta charset="utf-8" />
+    <meta charset="utf-8"/>
     <meta
       name="viewport"
       content="width=device-width, initial-scale=1, shrink-to-fit=no"
-    />
+/>
```
yndajas, line 8: Is this Prettier?

src/_layouts/default.html
```diff
@@ -55,5 +31,8 @@
       {% include page-content.html %}
     </main>
     {% include contribute.html %}
+    <script src="/assets/js/plausible.js">
+      plausible('pageview', {u: prepareUrl([])})
+    </script>
```
yndajas: Does this script need to execute at the end of the file rather than in the `head`?

src/_webpack/plausible.js
```diff
@@ -0,0 +1,20 @@
+// define the `plausible` function to manually trigger events
+window.plausible = window.plausible || function () {
+  (window.plausible.q = window.plausible.q || []).push(arguments)
+}
+
+export function prepareUrl (params) {
+  console.log('got here')
```
yndajas: Was this `console.log` meant to be deleted?

src/_layouts/default.html
```diff
@@ -21,8 +21,32 @@
...
+      function prepareUrl(params) {
+        const url = new URL(location.href)
+        const queryParams = new URLSearchParams(location.search)
+        let customUrl = url.protocol + "//" + url.hostname + url
+          .pathname
+          .replace(/\/$/, '')
+        for (const paramName of params) {
+          const paramValue = queryParams.get(paramName)
+          if (paramValue) 
+            customUrl = customUrl + '/' + paramValue
```
yndajas: I think you fixed this in the next commit, but it might be good for it to be fixed here to keep the commit history green

```javascript
          if (paramValue) {
            customUrl = customUrl + '/' + paramValue
          }
```

src/_layouts/default.html
```diff
@@ -31,8 +31,6 @@
       {% include page-content.html %}
     </main>
     {% include contribute.html %}
-    <script src="/assets/js/plausible.js">
-      plausible('pageview', {u: prepareUrl([])})
-    </script>
+    <script src="/assets/js/plausible.js"></script>
```
yndajas: Ah - I was wondering about this in the previous commit. Does it need to be a separate commit or could it just be squashed into the last?

src/_webpack/plausible.js
```diff
@@ -1,16 +1,21 @@
...
+}
+
+const prepareUrl = (paramsToRegister) => {
   const url = new URL(window.location.href)
   const queryParams = new URLSearchParams(window.location.search)
   let customUrl = url.protocol + '//' + url.hostname + url
     .pathname
     .replace(/\/$/, '')
-  for (const paramName of params) {
+  for (const paramName of paramsToRegister) {
```
yndajas: Is it worth amending the `captureAllPageViews` parameter to match this?

yndajas (approved):
I don't know how Plausible works (trust you and Adam on that!), but have a few quick tidy up suggestions

src/_layouts/default.html
lookupdaily: It may be.. I don't think we have any project rules for html...

yndajas: Yeah probably not - we do have a `.prettierignore` file but I can't remember what `**` means. Everything?

PR #1209, opened 2023-10-10
Simplify principle description

johnwaterworth:
One final tweak, hopefully, to simplify the descriptions of the principles we've been working on.

Note that the proposal to improve these two principles came from a recent user researcher away day.

### DIFF
--- a/src/user-research/index.md
+++ b/src/user-research/index.md
@@ -17,7 +17,7 @@ Our principles are not rules. They guide our work, keep us improving as a team, 
 
 1. **Help teams understand and value people**
 
-   Researchers at dxw help our teams to build a deep understanding of the people who use our services, and to value and act on that understanding. That’s how we contribute to good decisions and public services that work well for the people who need them.
+   Researchers at dxw help our teams to build a deep understanding of the people who use our services, and to value and act on that understanding. So our teams can make good decisions and create public services that work well for the people who need them.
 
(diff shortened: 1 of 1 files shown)


src/user-research/index.md
```diff
@@ -17,7 +17,7 @@ Our principles are not rules. They guide our work, keep us improving as a team,
 
 1. **Help teams understand and value people**
 
-   Researchers at dxw help our teams to build a deep understanding of the people who use our services, and to value and act on that understanding. That’s how we contribute to good decisions and public services that work well for the people who need them.
+   Researchers at dxw help our teams to build a deep understanding of the people who use our services, and to value and act on that understanding. So our teams can make good decisions and create public services that work well for the people who need them.
```
yndajas, line 20: This one seems a little more jarring than the `And`-starting sentences to me. Perhaps because it's more common to start with `And` than `So`

johnwaterworth: Let's just go back to the sentence we had before. I only changed it to avoid saying 'help our teams' twice in a paragraph.

"We know that’s the best way to help teams make good decisions and create public services that work well for the people who need them."

johnwaterworth: The substantive change was to add ", and to value and act on that understanding." And everyone's happy with that.

yndajas (approved):

Issue #1192, opened 2023-08-21
Add formatter for html


lookupdaily:
Currently we have no formatter in place for html files. We could add Prettier for this, but would need to make sure it does not conflict with JavaScript Standard

This would help us to keep our code consistent when editing html files, and reduce formatting changes on changes to existing files (if a developer uses a formatter locally).

yndajas:
Could we just tell it to ignore `.js` files? Or alternatively switch to Prettier wholesale and drop Standard?

lookupdaily:
Yep, that's how we've done it on our current project (ignoring js files). If we binned standard, we'd need to use eslint anyway, as they have [separate concerns](https://prettier.io/docs/en/comparison), however standard covers formatting as well as rule checking i believe?  

Prettier has [guidance on integrating with eslint](https://prettier.io/docs/en/integrating-with-linters), but not standard.

PR #1212, opened 2023-10-16
Add Stylelint

yndajas:
... with rules that order declarations, so that any future changes are devoid of arbitrary reorganisation, instead focused on meaningful change

Per [this client project commit][1], `stylelint-selector bem-pattern` is added and `selector-class-pattern` is turned off as by default Stylelint (or the standard rules) require strictly kebab case class names

[1]: https://github.com/ministryofjustice/hmpps-accredited-programmes-ui/commit/025ed66b635efdf606dbcea37fa0a42f0b42b6c7

### DIFF
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -1,0 +1,10 @@
+{
+  "extends": ["stylelint-config-standard-scss", "stylelint-config-clean-order"],
+  "plugins": ["stylelint-selector-bem-pattern"],
+  "rules": {
+    "plugin/selector-bem-pattern": {
+      "preset": "bem"
+    },
+    "selector-class-pattern": null
+  }
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,6 +15,10 @@
         "babel-loader": "^9.1.2",
         "normalize.css": "8.0.1",
         "standard": "^17.0.0",
+        "stylelint": "^15.10.3",
+        "stylelint-config-clean-order": "^5.2.0",
+        "stylelint-config-standard-scss": "^11.0.0",
+        "stylelint-selector-bem-pattern": "^3.0.1",
         "webpack": "^5.76.0",
         "webpack-cli": "^5.0.1"
       }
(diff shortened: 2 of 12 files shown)


lookupdaily (approved):
This looks good, thanks for sorting. Do we want to do any formatting checks on html and css?

yndajas:
> This looks good, thanks for sorting. Do we want to do any formatting checks on html and css?

I think Stylelint does the CSS? Something on the HTML might be useful!

lookupdaily:
> I think Stylelint does the CSS? Something on the HTML might be useful!

As of [Stylelint v15](https://stylelint.io/migration-guide/to-15#deprecated-stylistic-rules) it doesn't. We encountered this [on our client project](https://github.com/DFE-Digital/find-information-about-academies-and-trusts/pull/102) when updating to v15 so have set up both stylelint and prettier

yndajas:
> > I think Stylelint does the CSS? Something on the HTML might be useful!
> 
> As of [Stylelint v15](https://stylelint.io/migration-guide/to-15#deprecated-stylistic-rules) it doesn't. We encountered this [on our client project](https://github.com/DFE-Digital/find-information-about-academies-and-trusts/pull/102) when updating to v15 so have set up both stylelint and prettier

😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱

`stylelint-config-clean-order` is doing some work here, but that explains why I was seeing some strange whitespace after running `--fix`!

Yeah, I think it might be good to add Prettier to deal with HTML and CSS formatting in that case, as long as they don't contradict each other in declaration order

We are telling Prettier to ignore everything at present, but maybe we can just tell it to ignore JavaScript (if we're retaining Standard)

yndajas:
@lookupdaily I guess #1192 covers it

PR #1238, opened 2023-12-04
Remove Strategy Unit content

yndajas:
... and the Strategy Unit section from the admin interface

@DKofie-00 do the changes to the dxw time and sales pages also make sense?


DKofie-00 (approved):
These changes look fine to me. Thanks, Ynda.

PR #1270, opened 2024-01-11
Link from accessibility principle to accessibility manual

johnwaterworth:
Link the principle 'Make everything we do accessible' to the accessibillty manual.

### DIFF
--- a/src/about-us/our-mission-values-and-principles.md
+++ b/src/about-us/our-mission-values-and-principles.md
@@ -1,8 +1,8 @@
 ---
 title: Our mission, values and principles
 related_order: 2
+last_reviewed_at: ""
 ---
-
 ## Our mission
 
 dxw's mission is to:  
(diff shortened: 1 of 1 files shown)


yndajas:
I'm a little out of the loop, but I'm not sure if we're ready to go public on the accessibility manual yet, so it might be best to hold off on this change for a while?

johnwaterworth:
Thanks Ynda. We can leave this one here for a while.

calumryan (approved):

src/about-us/our-mission-values-and-principles.md
```diff
@@ -124,15 +124,14 @@ However, there are some things that we must keep private.
 
 ### Make everything we do accessible
 
-At dxw we want to make sure that everyone can access the digital services we
+At dxw we want to make sure that everyone can access the services we
```
ZuleimaMR, line 127: why not products and services?
why just services?

johnwaterworth: Because we are usually talking about public services or government services. Some clients do use the word 'product' to describe the large parts that make up a service. But nobody talks about 'public products'. And GDS call Notify, Pay, Forms and One Login 'platforms' or 'tools' or 'technologies' rather than products.

ZuleimaMR: thank you for explaining this

PR #1279, opened 2024-01-25
Add a section on guiding participants to support

johnwaterworth:
Add a new section that gives examples of the kind of support we might recommend.

### DIFF
--- a/src/user-research/doing-research-safely.md
+++ b/src/user-research/doing-research-safely.md
@@ -17,6 +17,24 @@ The Service Manual has a good guide to
 [researching emotionally sensitive subjects](https://www.gov.uk/service-manual/user-research/researching-emotionally-sensitive-subjects),
 that covers preparing for, running and dealing with research.
 
+## Guiding participants to appropriate support
+
+Once we have built a rapport with a participant, they may ask us questions, or want to talk about things that are beyond our
+competency as user researchers.
+
+These can be related to the subject of the research. But can be about suicide, trauma, discrimination, ableism etc.
+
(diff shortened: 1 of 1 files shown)


src/user-research/doing-research-safely.md
```diff
@@ -24,7 +24,7 @@ competency as user researchers.
 
 These can be related to the subject of the research. But can be about suicide, trauma, discrimination, ableism etc.
 
-While we are always sympathetic and not judgemental, we do not try to offer advice. Instead we guide people to appropriate support.
+While we are always sympathetic and not judgemental, we do not try to offer advice. Instead, we guide people to appropriate support.
```
ZuleimaMR, line 27: Being sympathetic and not judgemental are only a few of the behaviours/competencies needed to do care-centred and trauma informed research. Hence, in the guide that I have written I added a section that refers to adding these two topics.

In an ideal situation (which would be for me having those section ready), I would prefer that this section in the playbook says: While we do care-centred and trauma informed research, we refrain from offering advise and instead guide people to appropriate support. 

I have material to build guidance on trauma informed and care centred research. I am thinking of introducing this in our next dxw community session which I am chairing.

ZuleimaMR: I will approve this, but i think that this will need to be revised after seeing how our community responds to introducing concepts of care-centred and trauma-informed practice.

ZuleimaMR (approved):

src/user-research/doing-research-safely.md
johnwaterworth: Thanks Zuleima. And yes, let's expand the guidance when we can.

PR #1284, opened 2024-01-26
Link to advice on making an access guide

johnwaterworth:
Link to the Sensory Trust's advice on making a visitor access guide.


ZuleimaMR (commented):
I am concerned that many of the hyperlinks that i had in the original document, guiding people to relevant sources have completly dissapeared from your adapted version John.

There are many, but this is one example where i was linking, for instance to paper on wayfinding, because it is not a common concept.

I do not agree on those hyperlinks being gone. Not only here.

johnwaterworth:
Thanks Zuleima, I've created a separate pull request to add those links.

ZuleimaMR (approved):

PR #1289, opened 2024-01-29
Update guidance for current practice

johnwaterworth:
Reference Google Meet rather than GoToMeeting as our preferred video call service.
Be consistent about we rather than you.
Re-order sections to cover the most common situations first.
Add more links to other guidance.

### DIFF
--- a/src/user-research/taking-notes-and-recording-research-sessions.md
+++ b/src/user-research/taking-notes-and-recording-research-sessions.md
@@ -1,7 +1,7 @@
 ---
 title: Taking notes and recording research sessions
 related_order: 180
-last_reviewed_at: ""
+last_reviewed_at: 2024-01-29T13:31:14.454Z
 ---
 To capture what happens in our research sessions we take notes and photos, make
 audio, video and screen recordings, and take copies of paperwork and other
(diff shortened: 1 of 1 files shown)


src/user-research/taking-notes-and-recording-research-sessions.md
```diff
@@ -21,116 +17,107 @@ Here we provide additional, specific guidance for researchers at dxw.
...
-* record video using an external camera and Quicktime
-* record your laptop screen using Quicktime or Screenshot
+* The participant may decide they don’t want to be recorded in a particular way, so we are prepared to switch from video to an audio-only recording, or to just take notes
+* Our preferred device, app or service may not be working properly, so we have an alternative like a paper notebook
+* Our note taker, assistant or observer may not be able to join the session, so we are prepared to take notes and recordings ourselves
 
-If research participants might use your laptop or see your screen during a
-research session, it’s a good idea to create a separate ‘research’ user account
-on your laptop with things like notifications and browser autofill turned off.
+If our participant refuses all recording, including note taking, we accept that we might need to abandon the session.
```
ZuleimaMR, line 56: Joh, this feels one sided to one scenario, are we then advised to not go ahead with a research session when there is not recording consent of any type?
I would say no, that we should go ahead with a session if we are experienced enough take quick notes and reflective notes after the session. For instance, at the Department for Education I had about 4 participants that refused recording of any type, there were not takers but no recording. I still went ahead with the session and it was useful for the research project.

ZuleimaMR (commented):
i left a comment about line 56, a small clarification needed.

src/user-research/taking-notes-and-recording-research-sessions.md
johnwaterworth: My intention was to give researchers permission to abandon a session, rather than to say they had to.
In the past I've seen researchers feel like that had to complete a session no matter what happened.

Would this be clearer:
If our participant refuses all recording, including note taking, we can continue with a session. But accept that we might need to adapt or abandon the session.

ZuleimaMR: If your intention is to give user researchers permission to abandon the session then just say so.
I think that trying to say something without saying it, is not helpful.
Here it is compounded because there are to messages: they one you are trying to say but not saying and the fact that some user researchers level of experience will mean it is best for them to run or not run the session.

This section is about taking notes and recording research sessions. Do you think that there is another section in the playbook where we can say or we are already saying explicitly that user researchers can opt to end the session for whatever reason (distress, competency when video/audio recording is not consented, other...)?

It would be clearer to say then: If our participant refuses all recording, including note takers, we might choose to continue with the session or cancel it.

Note how I have referred explicitly to including 'note takers' and not taking notes, because, are we introducing a research step in which we ask participants if we as user researchers can take notes APART from having note takers?

johnwaterworth: I have meetings now. I'll think about those comments, and respond tomorrow.

johnwaterworth: Following our discussion, I've changed this to:

We can decide to cancel a session if essential note taking, recording or other data collection is not possible. Whether because of technical problems, or because the participant does not consent.

ZuleimaMR (approved):

PR #1296, opened 2024-02-07
Work in progress - restructure work we do

DKofie-00:
Testing - not ready for approval yet


src/design/content-design/index.md
```diff
@@ -0,0 +1,39 @@
+---
+title: Content design at dxw
```
yndajas, line 2: This should have `related_order` and `last_reviewed_at`. Same on a few other pages

src/design/service-design/service-design-approach.md
```diff
@@ -1,9 +1,8 @@
 ---
 title: Service design approach
-redirect_from:
-  - /guides/service-design-at-dxw
-last_reviewed_at: ""
```
yndajas, line 5: `last_reviewed_at` should stay and the `redirect_from` should move to the new index

src/work-we-do/providing-techical-support.md
```diff
@@ -0,0 +1,59 @@
+---
+title: Providing technical support
+related_order: 3
+last_reviewed_at: ""
+---
+
+The technology team supports some of our client services. These are live services, usually (but not always) built by dxw, for which we provide ongoing support with hosting, maintenance, or bug fixes. We split our support into 2 units: [GovPress](https://playbook.dxw.com/govpress-unit/support/) and Delivery+.
```
yndajas, line 7: Playbook links need to be relative (I think we might need to fix a few on other pages separately)

src/sales-and-marketing/index.md
```diff
@@ -1,10 +1,60 @@
...
+
+
+
+
+
+---
+title: Sales and marketing team principles
+related_order: 1
+last_reviewed_at: ""
+---
```
yndajas, line 57: Looks like this and what follows has been moved to a separate page so it should be deleted from here

src/tech/development-and-technical-operations-team principles.md
```diff
@@ -0,0 +1,201 @@
+---
```
yndajas, line 1: The space in the filename needs to be a dash

yndajas (commented):
Spotted a few small things to fix from a quick scan. I can do some of this if that helps!

yndajas (approved):
I haven't looked at this on a line-by-line or commit-by-commit basis, but from scanning the overall diff it looks like the only content that's removed here (rather than reorganised) is the first line support section in "Support and on call". I've checked with the tech ops team about this removal and there haven't been any objections, and my understanding is that the content no longer represents how support works

PR #1302, opened 2024-03-05
Describe how *we* create and use discussion guides

johnwaterworth:
Describe how we create and use discussion guides, not how you should do that. To be consistent with the style of our other guides.


ZuleimaMR (approved):
I do not have any suggestions for changes or improvements.

PR #1308, opened 2024-03-20
Add notes on practicalities when the sole technologist on a project

lozette:
Add a section about practicalities with the commit/review/merge cycle when a developer is the sole technologist on a project.

### DIFF
--- a/src/tech/how-we-do-development.md
+++ b/src/tech/how-we-do-development.md
@@ -412,3 +412,40 @@ maintainers and contributors understand the reasoning for not doing so.
 See
 [Tech Team RFC-019](https://github.com/dxw/tech-team-rfcs/blob/main/ways-of-working/rfc-019-use-changelogs-to-track-changes.md)
 for more context around this decision.
+
+## Developing alone
+
+Developers at senior level and above are expected to be able to be the sole
+technologist on a project. This can uncover issues with the review/merge cycle,
+especially if the project is beyond the discovery or alpha phase, as there is
+no developer to collaborate with and review pull requests.
+
+### Raise the issue as a project risk
+
+Generally pull requests require an approving review to be merged. Without another
+developer on the project who knows the domain, this can result in a blocker to
+development. The client should be made aware that this will be a sticking point
+and may slow the development process. The risk should be recorded in writing.
+
+### Work with the client to find solutions
+
+The client may have other developers or developer-adjacent people who know the
+product landscape. 
+
+If the repository belongs to the client, another developer or technologist can
+be given write access to the repository and the ability to approve pull requests.
+If the repository belongs to dxw, they can be added to the repository as a guest.
+
+### Ask dxw developers on other projects for help
+
+If the client does not have anyone suitable, an external (not on the project) dxw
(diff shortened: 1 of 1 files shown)


src/tech/how-we-do-development.md
```diff
@@ -411,4 +411,24 @@ maintainers and contributors understand the reasoning for not doing so.
 
 See
 [Tech Team RFC-019](https://github.com/dxw/tech-team-rfcs/blob/main/ways-of-working/rfc-019-use-changelogs-to-track-changes.md)
-for more context around this decision.
\ No newline at end of file
+for more context around this decision.
+
+## Developing alone
+
+Developers at senior level and above are expected to be able to be the sole technologist on a project. This can uncover issues with the commit/review/merge cycle, especially if the project is beyond the discovery or alpha phase, as there is no developer to collaborate with and review pull requests.
```
yndajas, line 418: "commit/review/merge cycle" -> this seems more focused on reviewing specifically. Are there impacts on commits and merges as well?

src/tech/how-we-do-development.md
```diff
@@ -411,4 +411,24 @@ maintainers and contributors understand the reasoning for not doing so.
...
+
+### Work with the client to find solutions
+
+The client may have other developers or developer-adjacent people who know the product landscape. 
+
+If the repository belongs to the client, this person can be given write access to the repository and the ability to approve pull requests. If the repository belongs to dxw, they can be added to the repository as a guest.
+
+### Ask dxw developers on other projects for help
+
+If the client does not have anyone suitable, a dxw developer can review pull requests alongside the solo developer. If the repository belongs to the client, the dxw developer should be granted access to merge pull requests, *or* the solo developer should be allowed to merge their own pull requests. The client should be made aware of this requirement, should their security policies not allow developers to merge their own pull requests.
```
yndajas, line 432: dxw developer and solo developer generally not being mutually exclusive descriptions confused me a little here. Small wording suggestion below to clarify that and something else:

```suggestion
If the client does not have anyone suitable, a dxw developer can review pull requests for the solo developer. If the repository belongs to the client, the reviewer should be granted access to merge pull requests, *or* the solo developer should be allowed to merge their own pull requests. The client should be made aware of this requirement, should their security policies not allow developers to merge their own pull requests.
```

Also:
- "the dxw developer should be granted access to merge pull requests" - is it more that they should be granted write access if the repository requires PR approval from someone with write access? I would've thought the person opening the PR would typically be responsible for merging regardless of the review process?
- "the solo developer should be allowed to merge their own pull requests" - without requiring approval? I'd guess merging their own pull requests in general is not a problem, it's just the approval part?

yndajas (approved):
Generally looks good to me, but a few suggestions

src/tech/how-we-do-development.md
lozette: There are impacts on merges when a client's security policy disallows developers merging their own commits (DfE does this for example). I'll take out "commit" though, you're right there.

lozette: You're right I'm getting confused with approvals with write access vs approvals with no write access 🤦

PR #1311, opened 2024-04-08
Update Playbook licence from CC BY-NC 2.0 to CC BY-NC 4.0

jdudley1123:
# Context
The Playbook is licensed with a Creative Commons Attribution-NonCommercial 2.0 UK: England & Wales licence. That licence has since been replaced by version 4.0, which is an international licence and not region specific.

[Creative Commons recommends](https://creativecommons.org/licenses/by-nc/2.0/uk/deed.en) that we use the newer version of the licence because it is "more user-friendly and more internationally robust".

# This PR
This PR changes the licence referenced in the readme from version 2.0 UK: England & Wales to version 4.0 International.

It also adds the licence as LICENSE file which GitHub can then pick up and highlight in the repository metadata.

# Note
If this PR is rejected and we stick with version 2.0, we need to update the link from https://creativecommons.org/licenses/by-nc/2.0/uk/ to https://creativecommons.org/licenses/by-nc/2.0/uk/deed.en, as the current link is broken, and we should add a LICENSE file for that licence.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,407 @@
+Attribution-NonCommercial 4.0 International
+
+=======================================================================
+
+Creative Commons Corporation ("Creative Commons") is not a law firm and
+does not provide legal services or legal advice. Distribution of
+Creative Commons public licenses does not create a lawyer-client or
+other relationship. Creative Commons makes its licenses and related
+information available on an "as-is" basis. Creative Commons gives no
+warranties regarding its licenses, any material licensed under their
+terms and conditions, or any related information. Creative Commons
+disclaims all liability for damages resulting from their use to the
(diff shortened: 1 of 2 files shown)


LICENSE
```diff
@@ -0,0 +1,407 @@
+Attribution-NonCommercial 4.0 International
```
yndajas, line 1: Do we definitely want a copy rather than just the link?

yndajas (approved):

LICENSE
jdudley1123: If we add a copy then GitHub adds it to the repo metadata like this
image
image

yndajas: Nice

PR #1315, opened 2024-05-07
Use dxw internal renovate config

jdudley1123:
We have a load of open dependency PRs on this repo. To automate some of this, I've switched us to using our internal config for Renovate so that it:
- Automatically merges minor updates, as well as patch
- Combines patch updates and minor updates into a single pull request

See https://github.com/dxw/renovate-config


yndajas (approved):
Go on then. Shall we merge [this one](https://github.com/dxw/rails-template/pull/672) too while we're at it?

PR #1324, opened 2024-05-29
Upgrade Ruby

yndajas:
This upgrades Ruby, in the process upgrading `sass-embedded` and restoring it as a direct dependency, which fixes a Netlify build failure

Closes #1313


snim2 (approved):
Preview looks good :)

PR #1457, opened 2024-12-24
Update `jekyll-last-modified-at` gem

lookupdaily:
When trying to update my local setup of the playbook, bundle install exits with the error:

```
An error occurred while installing posix-spawn (0.3.15), and Bundler cannot continue.

In Gemfile:
  jekyll-last-modified-at was resolved to 1.3.0, which depends on
    posix-spawn
```

The `jekyll-last-modified-at` gem has since been modified to replace this dependency, so updating the gem resolves the error. (https://github.com/gjtorikian/jekyll-last-modified-at/issues/97#issuecomment-2179282124)


lookupdaily:
Not an issue - I hadn't spotted that my pull from main had failed, so was running on an old version 😬

PR #1460, opened 2024-12-24
Update authenticated hrefs to exclude in html proofer

lookupdaily:
We currently use the [html-proofer](https://github.com/gjtorikian/html-proofer) plugin to help us catch broken links in the playbook. 

Some external links in the playbook are authenticated due to privacy level, or will be directed to an authentication screen for other reasons (e.g. LinkedIn). Updating the list to include more domains that currently fail the script for this reason.

Replacing the dxw private repository links with a catch-all for the dxw GitHub organisation, as there quite a few links to private repositories and adding them individually will be harder to maintain.

### DIFF
--- a/script/utils/check-html
+++ b/script/utils/check-html
@@ -23,9 +23,10 @@ IGNORE_HTTP_STATUS="200"
 #   https://github.com/jekyll/jekyll/blob/master/script/proof
 IGNORE_AUTHENTICATED_HREFS=$(ruby -e 'puts %w{
   expenses.xero.com/!wrUP-/detail/create-new/
-  github.com/dxw/govpress-developer-docs
-  github.com/dxw/judiciary-middleware
-  github.com/dxw/mind-side-by-side
+  github.com/dxw
+  docs.google.com
+  linkedin.com
+  git.govpress.com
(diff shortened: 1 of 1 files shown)


lookupdaily:
mind.org.uk is currently failing the proofer with a 403 code, but loads fine in the browser. We may want to look at adding this to the `IGNORE_ANY_OTHER_FALSE_POSITIVE_HREFS` list

script/utils/check-html
```diff
@@ -23,9 +23,10 @@ IGNORE_HTTP_STATUS="200"
 #   https://github.com/jekyll/jekyll/blob/master/script/proof
 IGNORE_AUTHENTICATED_HREFS=$(ruby -e 'puts %w{
   expenses.xero.com/!wrUP-/detail/create-new/
-  github.com/dxw/govpress-developer-docs
-  github.com/dxw/judiciary-middleware
-  github.com/dxw/mind-side-by-side
+  github.com/dxw
+  docs.google.com
+  linkedin.com
+  git.govpress.com
```
rjw1, line 29: While this wont stop me approving this PR and merging. We have actually turned off git.govpress.com.

rjw1 (approved):

PR #1464, opened 2024-12-27
Update delegated authorities

rjw1:


snim2 (approved):
@rjw1 this needs bringing up to date with the target branch but it looks fine to me

PR #1503, opened 2025-03-06
Add contributing collection

snim2:
So we can edit this folder in the UI.

### DIFF
--- a/src/admin/config.yml
+++ b/src/admin/config.yml
@@ -109,6 +109,12 @@ collections:
     folder: src/commercial-ops
 
   - <<: *collection_defaults
+    name: contributing
+    label: Contributing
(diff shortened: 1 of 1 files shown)


rjw1 (approved):

src/admin/config.yml
```diff
@@ -108,6 +108,12 @@ collections:
     label_singular: Commercial Operations page
     folder: src/commercial-ops
 
+  - <<: *collection_defaults
+    name: contributing
+    label: Contributing
+    label_singular: Contributing
```
yndajas, line 114: I think this should be a "Contributing page"

snim2: Will raise a PR

PR #1569, opened 2025-05-19
Fix pages not appearing in menu

yndajas:
These pages don't appear in the menu because of having apostrophes in their filenames and/or titles (haven't quite gotten to the bottom of the problem). I tried changing the character from `’` to `'` but that didn't fix it. The easiest solution seems to be avoiding the contraction altogether, which also brings these pages in line with the sibling "If you are the secondary carer" page

### DIFF
--- a/src/staff-handbook/policies-and-procedures/parental-leave/if-you-are-adopting-or-having-a-baby-through-surrogacy.md
+++ b/src/staff-handbook/policies-and-procedures/parental-leave/if-you-are-adopting-or-having-a-baby-through-surrogacy.md
@@ -1,5 +1,5 @@
 ---
-title: If you’re adopting or having a baby through surrogacy
+title: If you are adopting or having a baby through surrogacy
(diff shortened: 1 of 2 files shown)


snim2 (approved):
OK. My understanding of your description is that this is not a Unicode issue, is that right?

yndajas:
> OK. My understanding of your description is that this is not a Unicode issue, is that right?

I'm not really sure what the issue is. The pages [exist](https://playbook.dxw.com/staff-handbook/policies-and-procedures/parental-leave/if-you%E2%80%99re-pregnant/) but I suspect the escaping is causing some issues when we're trying to [include them as child pages](https://github.com/dxw/playbook/blob/b8af08abdbf7996b9869ad861e750558dd1ce8f9/src/_includes/related_pages.html). I tried removing some of the processing and conditional logic in that template, but couldn't get the pages to show up

PR #1585, opened 2025-07-01
Add canonical links to all pages

snim2:


.node-version
```diff
@@ -1 +1 @@
-22.17.0
+22.5.1
```
yndajas, line 1: Guessing this was unintentional?

src/_layouts/default.html
```diff
@@ -6,6 +6,7 @@
       name="viewport"
       content="width=device-width, initial-scale=1, shrink-to-fit=no"
 />
+    <link rel="canonical" href="{{site.url}}{{page.url}}" />
```
yndajas, line 9: Might be good to explain the benefit of these in the commit message?

PR #1586, opened 2025-07-02
[spike] Add custom plugin to generate Netlify _redirects file

jacksonj04:
At the moment, the [jekyll-redirect-from plugin](https://github.com/jekyll/jekyll-redirect-from) relies on JavaScript to perform page redirects (since Jekyll cannot guarantee that where it's being deployed can properly reply with HTTP redirects). In our case, however, we're deploying to Netlify, which [_does_ support redirecting](https://docs.netlify.com/routing/redirects/).

This commit adds a new custom plugin which reads the `redirect_from` frontmatter of a page (same as the jekyll-redirect-from syntax) and generates an [`_redirects` file](https://docs.netlify.com/routing/redirects/#syntax-for-the-redirects-file) compatible with Netlify, eg:

```
/guides/agile-delivery-phases /work-we-do/how-we-deliver-work/agile-delivery-phases/ 301
/work-we-do/building-services/agile-delivery-phases /work-we-do/how-we-deliver-work/agile-delivery-phases/ 301
/guides/being-a-tech-lead-at-dxw /tech/being-a-tech-lead-at-dxw/ 301
...
```

## Next steps

This is a quick spike, and probably shouldn't be considered production ready. In particular:

- [ ] This needs testing in an actual Netlify deployment to make sure it behaves
- [ ] We'll need to do something to account for the existing plugin, probably either remove it (preferable to reduce noise) or stick a `!` on the redirect codes being generated so they happen forcibly even if the page already exists


src/contributing/index.md
```diff
@@ -130,29 +130,34 @@ order value set in `_config.yml` in the root of the project.
...
+Where you want to redirect a fragment within an existing page (eg `old_page#section`, where `section` now has its own dedicated page) you can use a JavaScript-powered solution. This should be the second choice where a HTTP redirect won't do the job, since it requires JavaScript to be enabled and has implications for search engines.
+
+To add one of these redirects, you will need to update `redirects.json`. This may need a developer's help. Please test this behaviour in the preview deployment before you merge the change.
 
 There is a redirect manager in the Netlify CMS. Within this admin section you can enter
 a `redirect from` and a `redirect to` value. This enables redirects (powered by
-`redirect.js`) which are not supported by the `jekyll-redirect-from` gem, such as
+`redirect.js`) which are not supported by Netlify's native redirects, such as
 redirecting what was previously a single section of a page to a new dedicated page. If
 you create an entry in this section please test it once the build has completed.
```
yndajas, line 158: **suggestion (non-blocking):** I would maybe restructure/reword this section to keep it accessible for non-technical contributors and avoid repeating the existing content, e.g.:
- the what, using less technical language e.g. "Where you want to redirect a section (or 'fragment') within an existing page"
- the how: using the redirect manager in Netlify CMS, noting that a developer _can_ help if needed
- example situations subsection
- how it works subsection (like you did for HTTP redirects), providing the technical stuff (JavaScript vs HTTP redirects, redirects.json, redirects.js, search engines...)

It might also be worth renaming the level four sections to e.g.:
- From an old page to a new page
- From a section of an old page to a new page

PR #1633, opened 2025-10-01
Remove deprecated CSS clip property

jacksonj04:
The [`clip` property](https://developer.mozilla.org/en-US/docs/Web/CSS/clip) has been deprecated for a while now, and its [replacement](https://developer.mozilla.org/en-US/docs/Web/CSS/clip-path) has been widely supported since 2020.

### DIFF
--- a/src/_sass/_accessibility.scss
+++ b/src/_sass/_accessibility.scss
@@ -11,6 +11,5 @@
 
   white-space: nowrap !important;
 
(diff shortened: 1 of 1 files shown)


yndajas (approved):
The one element that uses this class is still visually hidden, so this seems fine. Maybe worth doing the same here: https://github.com/dxw/accessibility-manual/blob/main/src/_sass/_accessibility.scss#L13

